PR #5, opened 2023-11-15
Update standards names in README.md

mpeg2tom:
Standards names should be "JPEG XS" (see: "ISO/IEC 21122-1:2022 JPEG XS low-latency lightweight image coding system" https://www.iso.org/standard/81551.html) and SMPTE ST 2110 (see: https://www.smpte.org/smpte-st-2110-faq)


qiang-han (approved):
Correct!

PR #8, opened 2023-11-16
Update the two samples with several fixes

djie1:
It's not one perfect fix. There are several issues left with marks like "TODO"


qiang-han (approved):
looks good to me.

PR #24, opened 2023-11-23
Update sdk and sample, to carry metadata and data length to receiver side

djie1:
mcm_buffer.metadata/len are supposed to be filled by user from sender side. The information will be sent to receiver side, and can be retrieved by application.


qiang-han (approved):
looks good to me.

jianbingzhu (commented):
looks good to me

PR #55, opened 2023-12-19
Support MSM Control Plane.

qiang-han:


jianbingzhu (approved):
just interface, has not implemented. No regression for current code

PR #67, opened 2023-12-27
add more number of requested memzone segments to set memzone_max manu…

jianbingzhu:
add more number of requested memzone segments to set memzone_max manually for 1K IP camera sessions


media-proxy/src/proxy_context.cc
```diff
@@ -200,6 +201,7 @@ void ProxyContext::ParseStInitParam(const mcm_conn_param* request, struct mtl_in
     st_param->rx_queues_cnt[MTL_PORT_P] = 32;
     st_param->tx_queues_cnt[MTL_PORT_P] = 32;
     st_param->lcores = NULL;
+    st_param->memzone_max = 3000;
```
qiang-han, line 204: miss the declaration of "memzone_max".

qiang-han:
need to rebase to the latest source code.

media-proxy/src/proxy_context.cc
jianbingzhu: The memzone_max has been declared in mtl_api.h and only need to update iMTL, which has been in https://github.com/OpenVisualCloud/Media-Transport-Library/pull/671.

qiang-han (approved):
LGTM

PR #70, opened 2024-01-02
[Refactor] Separate memif functions into modules

qiang-han:
This commit refactors the Memif source code for the media proxy by separating Memif functions into different modules according to the transport media protocol.


djie1:
are there changes in 1) definitions in header files? 2) function implementations?

jianbingzhu (approved):
LGTM

PR #120, opened 2024-05-28
High impact Coverity scan fixes

Mionsz:
High impact Coverity scan fixes that include:

- Buffer not NULL terminated, change `strncpy` method to safe equivalent `strlcpy`.
- Leaks of memory/pointers to system resources (CWE-404) fixed.
- Fixed functions with missing return values in some branches.
- Fixed the uses of an uninitialized variables (CWE-457)
- Minor other improvements and fixes

### DIFF
--- a/media-proxy/src/api_server_tcp.cc
+++ b/media-proxy/src/api_server_tcp.cc
@@ -288,6 +288,7 @@ void RunTCPServer(ProxyContext* ctx)
     address.sin_port = htons(port);
     if (bind(sock, (struct sockaddr*)&address, sizeof(address)) < 0) {
         fprintf(stderr, "error: cannot bind socket to port %d: %s\n", port, strerror(errno));
+        close(sock);
         return;
     }
 
--- a/media-proxy/src/api_server_tcp.cc
+++ b/media-proxy/src/api_server_tcp.cc
@@ -266,8 +266,6 @@ void RunTCPServer(ProxyContext* ctx)
     connection_t* connection = NULL;
     control_context* ctl_ctx = NULL;
     pthread_t thread;
-    size_t connection_t_size = sizeof(connection_t);
-    size_t control_context_size = sizeof(control_context);
 
     port = ctx->getTCPListenPort();
     if (port <= 0) {
@@ -303,21 +301,21 @@ void RunTCPServer(ProxyContext* ctx)
 
     while (keepRunning) {
         /* accept incoming connections */
-        connection = (connection_t*)malloc(connection_t_size);
+        connection = (connection_t*)malloc(sizeof(connection_t));
         if (!connection) continue;
 
-        memset(connection, 0x0, connection_t_size);
+        memset(connection, 0x0, sizeof(connection_t));
         connection->sock = accept(sock, &connection->address, (socklen_t*)&connection->addr_len);
         if (connection->sock <= 0) {
             free(connection);
         } else {
             /* start a new thread but do not wait for it */
-            ctl_ctx = (control_context*)malloc(control_context_size);
+            ctl_ctx = (control_context*)malloc(sizeof(control_context));
             if (!ctl_ctx) {
                 free(connection);
                 continue;
             }
-            memset(ctl_ctx, 0x0, control_context_size);
+            memset(ctl_ctx, 0x0, sizeof(control_context));
             ctl_ctx->proxy_ctx = ctx;
             ctl_ctx->conn = connection;
             if (pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx) == 0) {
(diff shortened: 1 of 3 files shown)


media-proxy/src/api_server_tcp.cc
```diff
@@ -295,22 +298,33 @@ void RunTCPServer(ProxyContext* ctx)
...
+            if(!ctl_ctx) {
+                free(connection);
+                continue;
+            }
+            memset(ctl_ctx, 0x0, control_context_size);
             ctl_ctx->proxy_ctx = ctx;
             ctl_ctx->conn = connection;
-            pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx);
-            pthread_detach(thread);
+            if(pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx) == 0) {
```
ko80, line 322:
```suggestion
            if (pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx) == 0) {
```

media-proxy/src/api_server_tcp.cc
```diff
@@ -295,22 +298,33 @@ void RunTCPServer(ProxyContext* ctx)
...
+
+        memset(connection, 0x0, connection_t_size);
         connection->sock = accept(sock, &connection->address, (socklen_t*)&connection->addr_len);
         if (connection->sock <= 0) {
             free(connection);
         } else {
             /* start a new thread but do not wait for it */
-            ctl_ctx = (control_context*)malloc(sizeof(control_context));
+            ctl_ctx = (control_context*)malloc(control_context_size);
+            if(!ctl_ctx) {
```
ko80, line 315:
```suggestion
            if (!ctl_ctx) {
```

media-proxy/src/api_server_tcp.cc
```diff
@@ -295,22 +298,33 @@ void RunTCPServer(ProxyContext* ctx)
...
-
     // signal(SIGINT, intHandler);
 
     while (keepRunning) {
         /* accept incoming connections */
-        connection = (connection_t*)malloc(sizeof(connection_t));
+        connection = (connection_t*)malloc(connection_t_size);
+        if (!connection) continue;
+
+        memset(connection, 0x0, connection_t_size);
```
ko80, line 308: Oh, what's the reason to create the separate `connection_t_size` variable for storing the size?

media-proxy/src/udp_h264_rx_sample.c
```diff
@@ -61,10 +63,9 @@ int rx_udp_h264_shm_init(rx_udp_h264_session_context_t* rx_ctx, memif_ops_t* mem
 
     /* unlink socket file */
     if (memif_ops->is_master && rx_ctx->memif_socket_args.path[0] != '@') {
-        struct stat st = { 0 };
-        if (stat("/run/mcm", &st) == -1) {
-            if (mkdir("/run/mcm", 0666) != 0) {
-                perror("Fail to create directory (/run/mcm) for MemIF.");
+        if (mkdir("/run/mcm", 0666) != 0) {
+            if(errno != EEXIST) {
```
ko80, line 67:
```suggestion
            if (errno != EEXIST) {
```

media-proxy/src/udp_h264_rx_sample.c
```diff
@@ -61,10 +63,9 @@ int rx_udp_h264_shm_init(rx_udp_h264_session_context_t* rx_ctx, memif_ops_t* mem
...
+                perror("Fail to access or create directory (/run/mcm) for MemIF.");
```
ko80, line 68:
```suggestion
                ERROR("Fail to access or create directory (/run/mcm) for MemIF.");
```

media-proxy/src/api_server_tcp.cc
```diff
@@ -274,7 +277,7 @@ void RunTCPServer(ProxyContext* ctx)
 
     /* create socket */
     sock = socket(AF_INET, SOCK_STREAM, 0);
-    if (sock <= 0) {
+    if (sock == -1) {
```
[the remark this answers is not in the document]
Mionsz: Theoretically documentation states that it should only return `-1` when socket was not created. But maybe you are right, this would be better idea.

media-proxy/src/api_server_tcp.cc
Mionsz: As this is being used inside a loop and in two separate places one after another so It is better to have this calculated.

ko80: Don't optimize prematurely (c)
I suggest not to introduce a separate variable to store the size but keep it as it was because seeing `sizeof(...)` makes it easy to read the code.

ko80 (approved):
Looks good to me

pwysock1-intel (approved):
Looks good.

PR #124, opened 2024-06-03
Fix high impact Coverity issues

ko80:
Fix high impact Coverity issues
* Fix dereference before null check in sample app
* Fix resource leakage in RunTCPServer()


pwysock1-intel (approved):
I do approve these changes.

PR #127, opened 2024-06-05
Cumulative Pre-Release Fixes Dockerfile oriented.

Mionsz:
### **THIS PR CONTAINS BREAKING CHANGES:**

[Common methods and scripting improved](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/e58edd8ac3ef46d05caddc2d706e2c6bf8b6b6ed) 

Common methods and scripting improved:
- added file `common.sh` in root repository directory with some of commonly used functions.
- added fail on error in every bash script, including pipe fail
- improved scripting by usage of absolute paths instead of relative ones (any directory removal with relative paths can have unexpected behavior)
- fixed bash script files with `#!/bin/bash` not in first line.

Also `build_docker.sh` file was added with basic workflow of building media-proxy Dockerfile with the same approach as mentioned above.
- build docker files using buildx toolkit
- read and pass possible proxy variables from environment

[Markdown and .gitignore files minor fixes](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/d8132eacfe3dde39c598a7496c679beafcbbf5a5) 

Markdown and `.gitignore` files minor fixes
- added proto files generated by `gRPC` to Git-Ignore
- added Microsoft VS-Code configuration files to Git-Ignore
- adjusted markdown files to GitHub standard
- minor fixes in readme text

[Header definition and msg_header simplification](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/08505bf4ff5436be47fd81b395472e0d9a7ada47) 

**THIS COMMIT CONTAINS BREAKING CHANGES:**
Header definition and `msg_header` simplification
- moved `magic_word` declaration outside to top-level include file HEADER_MAGIC_WORD
- changed type definition for magic word from `char` to `uint32_t` as for up to 4 letters this simplifies the flow.
- proposed workflow is compliant with the C++11 and all newer standards.
- casting example: `uint32_t res = *(uint32_t*)"MCM"`

[Debug added and minor code simplifications](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/0618d001069db4440ffe35752b0838dceb8f6605) 

Debug added and minor code simplifications
- added definition for DEBUG printing function
- fixed SDK logger header format issue
- fixed PROTO_MEMIF case - removed `;` after colon.
- extended debug information by adding methods name with variables
- removed infinity loops that could occur when waiting for non null variable like `deviceHandle`

[Partial Migration to C++17 for class ProxyContext](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/0f14002ad4bb38d0ccb0ceb61ccbfe1b031b85c5) 

Partial migration to C++17 for class `ProxyContext`
- changed all std::string copy methods to `std::string_view` (no copy parameter)
- removed old-fashioned ineffective methods for string operations from class file and constructor
- removed try{} catch{} from `ProxyContext` constructor as it could cause undefined behavior. Application should simply fail when main class constructor can not be initialized.
- added simple signal handling for the `TcpServer` for it to close threads when asked.
- redesigned main loop for `TcpServer` in more non-fail oriented way to allow for easy loop break when needed.

[Major Dockerfile changes and IMTL update](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/88e2dc824e090202920a369fe9482bb6c0190ed1) 

Major Dockerfile changes and IMTL update:
Due to workflow and default values changes in the IMTL workflow, the `Dockerfile` had to be adopted. This includes fixes to DPDK drivers build process, IMTL build and usage of XDP repository.

Also minor changes like syntax versioning usage were also included:
`# syntax=docker/dockerfile:1`

[[msg_loop] Thread clean-up on exit](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/127/commits/ecf5483be25730d85a3dce336b263e0df60f2e55)

[msg_loop] Thread clean-up on exit.
On msg_loop exit, the thread that is exiting tries to do a cleanup based on session_id (value moved to the method scope so can be reused). This includes going through all `mtl_session_context_t` for running context (server) and checking the id.

Minor updates to common.sh:
- Removed non global text from Intel animated bash logo
- Made the logo little smaller and left only the one-liner description `Intel-Corporation`
- Adjustment to render and animation speed.

To use the animated logo, user should just type in his bash console:
(third line is an example, with all possible params (2), first is number of full sequences and second is a delay between steps:
```bash
. common.sh
print_logo_anim
print_logo_anim 3 0.025
```

[mSessionCount multisession bug fix](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/127/commits/c024701a2c0f27f8876a0b6edfbf689040984a1b) 

`mSessionCount` multisession bug fix:
- moved `mSessionCount` and mutex to private part of `ProxyContext` class.
- added method for safe increment of mSessionCount value.
- minor fix to constructor - `mTcpCtrlPort` condition.
- minor name refactor for `memif_ops_t`.
- explicit copy and assign constructor deletion for `ProxyContext`.

Signed-off-by: Milosz Linkiewicz <milosz.linkiewicz@intel.com>
@[Mionsz](https://github.com/OpenVisualCloud/Media-Communications-Mesh/commits?author=Mionsz)

### DIFF
--- a/media-proxy/src/api_server_tcp.cc
+++ b/media-proxy/src/api_server_tcp.cc
@@ -251,10 +251,17 @@ void* msg_loop(void* ptr)
     pthread_exit(0);
 }
 
-// void intHandler(int dummy)
-// {
-//     keepRunning = 0;
-// }
+void handleSignals(int sig_num)
+{
+    keepRunning = false;
+}
+
+void registerSignals()
+{
+    signal(SIGINT, handleSignals);
+    signal(SIGTERM, handleSignals);
+    signal(SIGKILL, handleSignals);
+}
 
 void RunTCPServer(ProxyContext* ctx)
 {
@@ -303,35 +310,35 @@ void RunTCPServer(ProxyContext* ctx)
     }
 
     INFO("TCP Server listening on %s", ctx->getTCPListenAddress().c_str());
-    // signal(SIGINT, intHandler);
+    registerSignals();
 
-    while (keepRunning) {
+    do {
         /* accept incoming connections */
         connection = (connection_t*)malloc(sizeof(connection_t));
-        if (!connection) continue;
-
-        memset(connection, 0x0, sizeof(connection_t));
-        connection->sock = accept(sock, &connection->address, (socklen_t*)&connection->addr_len);
-        if (connection->sock <= 0) {
-            free(connection);
-        } else {
-            /* start a new thread but do not wait for it */
-            ctl_ctx = (control_context*)malloc(sizeof(control_context));
-            if (!ctl_ctx) {
-                free(connection);
-                continue;
-            }
-            memset(ctl_ctx, 0x0, sizeof(control_context));
-            ctl_ctx->proxy_ctx = ctx;
-            ctl_ctx->conn = connection;
-            if (pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx) == 0) {
-                pthread_detach(thread);
+        if (connection) {
+            memset(connection, 0x0, sizeof(connection_t));
+            connection->sock = accept(sock, &connection->address, (socklen_t*)&connection->addr_len);
+            if (connection->sock > 0) {
+                /* start a new thread but do not wait for it */
+                ctl_ctx = (control_context*)malloc(sizeof(control_context));
+                if (ctl_ctx) {
+                    memset(ctl_ctx, 0x0, sizeof(control_context));
+                    ctl_ctx->proxy_ctx = ctx;
+                    ctl_ctx->conn = connection;
+                    if (pthread_create(&thread, 0, msg_loop, (void*)ctl_ctx) == 0) {
+                        pthread_detach(thread);
+                    } else {
+                        free(connection);
+                        free(ctl_ctx);
+                    }
+                } else {
+                    free(connection);
+                }
             } else {
                 free(connection);
-                free(ctl_ctx);
             }
         }
-    }
+    } while (keepRunning);
 
     INFO("TCP Server Quit: %s", ctx->getTCPListenAddress().c_str());
 
--- a/media-proxy/src/api_server_tcp.cc
+++ b/media-proxy/src/api_server_tcp.cc
@@ -1,19 +1,15 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2023 Intel Corporation
- *
  * SPDX-License-Identifier: BSD-3-Clause
- */
+*/
 
-// #include <signal.h>
 #include <bsd/string.h>
-
 #include "api_server_tcp.h"
 #include <mcm_dp.h>
 
 static volatile bool keepRunning = true;
 
-typedef struct
-{
+typedef struct {
     int sock;
     struct sockaddr address;
     int addr_len;
@@ -27,18 +23,20 @@ typedef struct _control_context {
 void* msg_loop(void* ptr)
 {
     int ret = 0;
-    char* buffer = NULL;
     int len = 0;
+    char* buffer = NULL;
+    bool sessionKeepRunning = true;
     control_context* ctl_ctx = NULL;
     connection_t* conn = NULL;
     ProxyContext* proxy_ctx = NULL;
     long addr = 0;
     mcm_proxy_ctl_msg msg = {};
+    uint32_t session_id = 0;
     uint32_t intf_id = 0;
 
     if (!ptr) {
+        ERROR("msg_loop(void* ptr): Illegal Parameter, ptr==NULL.");
         pthread_exit(0);
-        INFO("Illegal Parameter.");
     }
 
     ctl_ctx = (control_context*)ptr;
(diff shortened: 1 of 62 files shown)


media-proxy/src/proxy_context.cc
```diff
@@ -25,49 +19,45 @@ ProxyContext::ProxyContext(void)
...
-    istringstream f(mTcpCtrlAddr);
-    string s;
-
-    while (getline(f, s, ':')) {
-        sub_str.push_back(s);
-    }
-
-    if (sub_str.size() != 2) {
-        INFO("Illegal TCP listen address.");
+ProxyContext::ProxyContext(std::string_view rpc_addr, std::string_view tcp_addr)
```
ko80, line 22: Have you changed the class from `std::string` to `std::string_view` only to be able to use it's methods in this constructor? I suggest not to change the classes everywhere but to use this class locally in the constructor only.

media-proxy/src/api_server_tcp.cc
```diff
@@ -73,38 +71,32 @@ void* msg_loop(void* ptr)
...
-                    int ret = read(conn->sock, buffer + bytesRead, msg.command.data_len - bytesRead);
-                    if (ret <= 0) {
-                        INFO("Failed to read command parameters.");
-                        free(buffer);
-                        buffer = NULL; // Reset the pointer to indicate failure
-                        break; // Exit the loop on failure
-                    }
-                    bytesRead += ret;
+                while (bytesRead < msg.command.data_len && ret >= 0) {
+                    ret = read(conn->sock, buffer + bytesRead, msg.command.data_len - bytesRead);
```
ko80, line 80: I suppose, if you're still using a blocking socket, there is a high probability that this `read` call will be blocked when a termination signal occurs. And you'll never get a graceful shutdown in this case.

media-proxy/src/proxy_context.cc
Mionsz: No. I have changed it to `std::string_view` as this is the proper and fastest wat of passing `std::string` since `C++17`.
I do not even mention that the constructors like for ex. copy constructor should be explicitly disabled and it is not. Here in the overloaded version of constructor the main thing that took my attention was using try(), catch() statement inside it. 
Such construct - even if made possible by the standard itself - should no be used when constructing main class of the whole application. I do not even mention how the reverse stack unwrapping in case of unhandled exception would go.
So to sum this part up, I removed the `try` and `catch` as main constructor (initialization) should be allowed to fail and propagate to the caller. I also changed `string` to `string_view` - the first one in this form always does hard copy of full string when passed and the second one is secure no-copy equivalent (mostly).

Mionsz: Also here is a simple speed comparison for both of them: 
image
[https://stackoverflow.com/a/40129046]

media-proxy/src/api_server_tcp.cc
Mionsz: You are right here - I had similar thoughts. But on the other hand, in this loop I have included `keepRunning` from the main loop and `sessionKeepRunning` so I am/was looking for a way to force this thread to be shut down in case the client have disconnected on the other end of socket and cleanup after everything but at the moment I am looking for your//other reviewer possible ideas.

media-proxy/src/proxy_context.cc
ko80: Agree. Thanks for explaining that.

Mionsz:
I have added to commits with some more fixes. Description have been updated to include those bellow:

[[msg_loop] Thread clean-up on exit](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/127/commits/ecf5483be25730d85a3dce336b263e0df60f2e55)

[msg_loop] Thread clean-up on exit.
On msg_loop exit, the thread that is exiting tries to do a cleanup based on session_id (value moved to the method scope so can be reused). This includes going through all mtl_session_context_t for running context (server) and checking the id.

Minor updates to common.sh:

Removed non global text from Intel animated bash logo
Made the logo little smaller and left only the one-liner description Intel-Corporation
Adjustment to render and animation speed.
To use the animated logo, user should just type in his bash console:
(third line is an example, with all possible params (2), first is number of full sequences and second is a delay between steps:

```
. common.sh
print_logo_anim
print_logo_anim 3 0.025
```

[mSessionCount multisession bug fix](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/127/commits/c024701a2c0f27f8876a0b6edfbf689040984a1b)

`mSessionCount` multisession bug fix:

- moved `mSessionCount` and mutex to private part of `ProxyContext` class.
- added method for safe increment of `mSessionCount` value.
- minor fix to constructor - `mTcpCtrlPort` condition.
- minor name refactor for `memif_ops_t`.
- explicit copy and assign constructor deletion for `ProxyContext`.

media-proxy/src/api_server_tcp.cc
```diff
@@ -1,19 +1,15 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2023 Intel Corporation
```
pwysock1-intel, line 2: Please update Copyright (c) 2023 Intel Corporation to Copyright (c) 2024 Intel Corporation

Mionsz:
Branch had to be rebased on to main.

soopel (approved):
Approved

PR #128, opened 2024-06-06
FFmpeg fix for default yuv422p10le pixel format

Mionsz:
FFmpeg fix for default yuv422p10le pixel format:

Follow-up change to recent, introduced in NAB2024, adjusted and merged changes to media-proxy. Main
change was introduction of RFC4175 compliant (aka. st2110-20 format) frame FMT PIX_FMT_YUV422P_10BIT_LE, which is now default output format for media-proxy.

Above requires adjustments to `FFmpeg` by introducing `getFrameSize()` method for `mcm_rx.c` part of implementation. For making this MTL independent, method was implemented as is in the code instead of being included for the framework.

Updated README.md to include possible shared libraries error where `ffmpeg` tries to load libraries from wrong path by default.

### DIFF
--- a/sdk/Dockerfile
+++ b/sdk/Dockerfile
@@ -1,78 +1,68 @@
+# syntax=docker/dockerfile:1
+
 # SPDX-FileCopyrightText: Copyright (c) 2023 Intel Corporation
-#
 # SPDX-License-Identifier: BSD-3-Clause
 
-# Use the specific version of ubuntu image
-FROM ubuntu:22.04@sha256:2b7412e6465c3c7fc5bb21d3e6f1917c167358449fecac8176c6e496e5c1f05f as builder
+ARG IMAGE_CACHE_REGISTRY=docker.io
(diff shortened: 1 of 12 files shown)


ko80 (approved):
Checked MCM FFmpeg plugin on HW. Looks good to me.

sdk/Dockerfile
```diff
@@ -1,78 +1,68 @@
...
-# Set the maintainer label
-LABEL maintainer="qiang.han@intel.com"
+ARG IMAGE_CACHE_REGISTRY
+ARG IMAGE_NAME
+FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME}
 
-# Set environment variables
-ENV DEBIAN_FRONTEND=noninteractive
+LABEL maintainer="milosz.linkiewicz@intel.com"
+LABEL org.opencontainers.image.title="Intel® MCM SDK" \
```
soopel, line 34: Need to use full Media Communication Mesh name instead of MCM

soopel (dismissed):
Approved

sdk/Dockerfile
Mionsz: Fixed all labels for Docker images.
Added licenses annotations.

soopel (approved):
Approved

PR #143, opened 2024-06-16
Fixed error: will always evaluate as 'false'.

Mionsz:
Fixed: error will always evaluate as 'false'. In ICMP_RESPONDER there was a check against static variable check against NULL 
Fixed: Two high impact issues, buffer not null terminated (BUFFER_SIZE) due to wrong size used. 
Some minor adjustments.


sdk/src/mcm_dp.c
```diff
@@ -30,7 +30,7 @@ static void parse_memif_param(mcm_conn_param* request, memif_socket_args_t* memi
     snprintf(memif_socket_args->app_name, sizeof(memif_socket_args->app_name), "memif_%s_%d", type_str, atoi(request->local_addr.port));
     snprintf(interface_name, sizeof(interface_name), "memif_%s_%d", type_str, atoi(request->local_addr.port));
     strlcpy((char*)memif_conn_args->interface_name, interface_name, sizeof(memif_conn_args->interface_name));
-    if (request->memif_interface.socket_path == NULL) {
+    if (strlen(request->memif_interface.socket_path) <= 1) {
```
ko80, line 33: Could you elaborate more on why you need this change?

ko80:
```suggestion
    if (!strlen(request->memif_interface.socket_path)) {
```
Maybe this snippet is better?

ko80 (approved):
Looks good to me

pwysock1-intel (approved):
Approved.

PR #150, opened 2024-06-20
Samples applications fix

Mionsz:
Samples applications fix.
Set default pixel format to PIX_FMT_YUV422P_10BIT_LE.


ko80:
I suppose, it's a workaround, not a fix of a segfault in media_proxy. Do you plan to modify media_proxy source code as well?

soopel (approved):
Approved

Mionsz:
Root cause of the problem is probably
https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/bca7af10080310c6b860a41cdc8824f7c0bb2aa8/media-proxy/src/mtl.c#L280

PR #152, opened 2024-06-25
Fix segfault in MCM FFmpeg plugin occurred in memcpy

ko80:
In mcm_read_packet() the length of the buffer received from media_proxy can be smaller than the configured frame size. Passing the size argument to memcpy() being larger than the source buffer size can lead to a segmentation fault.

This commit fixes this issue by passing the size argument to memcpy() as the lowest value between the configured frame size and the actual Rx buffer length.


Mionsz (approved):

soopel (approved):
Approved

PR #155, opened 2024-06-28
Update CMakeLists.txt

tszumski:
Update CMakeLists.txt append "-fPIC" flag, without that some compilers return error
```
/usr/bin/ld: CMakeFiles/sender_app.dir/sender_app.c.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: CMakeFiles/recver_app.dir/recver_app.c.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
collect2: error: ld returned 1 exit status
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/sdk/CMakeLists.txt
+++ b/sdk/CMakeLists.txt
@@ -26,6 +26,8 @@ include(compile_options)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
 set(CMAKE_INSTALL_IPATH ${CMAKE_INSTALL_PREFIX}/include)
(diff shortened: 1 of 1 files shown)


Mionsz (changes requested):
Could you add the same C  flags for the `memif` library, as it is also build as dynamically linked library. Probably in:
`sdk/3rdparty/libmemif/src/CMakeLists.txt`

tszumski:
> Could you add the same C flags for the `memif` library, as it is also build as dynamically linked library. Probably in: `sdk/3rdparty/libmemif/src/CMakeLists.txt`

Done

Mionsz (approved):

MateuszGrabuszynski (approved):
lgtm

Mionsz:
@pwysock1-intel @soopel can you check this one and approve if it is ok so that we don't stall?

soopel (approved):
Approved

PR #158, opened 2024-07-01
Full mcm library version control tag based.

Mionsz:
Full mcm library version control tag based.

- Libraries libraisr and libmcm_dp should use position code independent properties from now on.
- Added scripting for version from git parsing (exactly from tag) or if this fails default one is being set.

Changes default patch actions for ffmpeg-plugin
- Using 'apply' with '--whitespaces=fix '.
- Basic data (name and e-mail) included in the patch itself (commiter full data).
- Updated project version in FFmpeg MCM to reflect SDK version from tag.


soopel (approved):
Looks good. Approved

PR #160, opened 2024-07-02
Fix integer overflow in mcm_send_buffer_udp

Mionsz:
Fix integer overflow in `mcm_send_buffer_udp`

Additional value check was added.


soopel (approved):
Approved

PR #161, opened 2024-07-02
Update udp_impl.c and fix the ssize_t overflow issue.

Mionsz:
Update `udp_impl.c` and fix the `ssize_t` overflow issue.


soopel (approved):
Approved

PR #162, opened 2024-07-02
Update udp_impl.c

Mionsz:
More fixes to update udp_impl.c

### DIFF
--- a/sdk/src/udp_impl.c
+++ b/sdk/src/udp_impl.c
@@ -116,34 +116,48 @@ int mcm_send_buffer_udp(void* conn_ctx, mcm_buffer* buf)
 {
     int ret = 0;
     udp_context* ctx = NULL;
-    void* data = NULL;
-    ssize_t n = 0, len = 0;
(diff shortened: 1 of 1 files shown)


sdk/src/udp_impl.c
```diff
@@ -116,34 +116,48 @@ int mcm_send_buffer_udp(void* conn_ctx, mcm_buffer* buf)
...
     }
-
     ctx = (udp_context*)conn_ctx;
-
     data = buf->data;
-    len = buf->len;
-    if(len < 0) {
+    len = (size_t)buf->len; // Cast to size_t, ensure that buf->len is validated before casting
+
+    if (buf->len < 0) {
```
soopel, line 131: Shouldn't we check len here instead of buf->len? Or opposite in line 136.

Mionsz: True and now it is resolved.

soopel (approved):
Looks good now.

PR #166, opened 2024-07-04
Update dependabot.yml

Mionsz:
Update `dependabot.yml` to make checks once a week. Also added two more subdirectories with dockerfiles and given exact directory for `github-actions`


soopel (approved):
Approved

PR #168, opened 2024-07-05
Dockerfile size shrink to 265MB + Mtl-Manager

Mionsz:
Dockerfile size shrink to 265MB and Mtl-Manager added by default to the container. This allows usage of one container with shared libraries for two purposes:

- MtlManager running
- Media-Proxy running

This approach guarantee that MtlMAnager is build and there is no need for external project know-how.

To lunch MtlManager just use bellow parameter:
`--entrypoint=/usr/local/bin/MtlManager`


soopel (approved):
Looks good. Approved

MateuszGrabuszynski (approved):
lgtm :)

PR #173, opened 2024-08-07
Remove metadata from sent buffer

Sakoram:
Video data was overwritten with metadata, which resulted in not enough data being sent

### DIFF
--- a/sdk/samples/recver_app.c
+++ b/sdk/samples/recver_app.c
@@ -353,8 +353,6 @@ int main(int argc, char** argv)
             }
             break;
         }
-        printf("INFO: buf->metadata.seq_num   = %u\n", buf->metadata.seq_num);
-        printf("INFO: buf->metadata.timestamp = %u\n", buf->metadata.timestamp);
         printf("INFO: buf->len = %ld frame size = %u\n", buf->len, frm_size);
 
         clock_gettime(CLOCK_REALTIME, &ts_recv);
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -11,9 +11,11 @@ SCRIPT_DIR="$(readlink -f "$(dirname -- "${BASH_SOURCE[0]}")")"
 BUILD_TYPE="${BUILD_TYPE:-Release}"
 INSTALL_PREFIX="${INSTALL_PREFIX:-/usr/local}"
 
+# TODO remove ENABLE_ZERO_COPY
 cmake -B "${SCRIPT_DIR}/out" \
     -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
+    -DENABLE_ZERO_COPY=OFF \
```
Mionsz, line 18: why disable zero copy?
Rest seems rather ok

Sakoram: It causes data corruption. I don't know if the fault is on the MCM side or MTL, but it is highly possible that the issue is in MTL. More investigation is needed, and disabling zero copy is only a temporary fix.

Mionsz (dismissed):

build.sh
Sakoram: I removed this part as it is fixed in different PR

Mionsz (approved):

PR #174, opened 2024-08-09
Ext_frame: Corrupt frame output fix

Mionsz:
Ext_frame: Corrupt frame output fix
- Frame size corrected
- Check for transport and frame size added
- MTL mem allignment added


soopel (approved):
Looks good to me.

Sakoram (approved):

Mionsz:
image
image

PR #175, opened 2024-08-09
Add skeleton for unit tests

Sakoram:


CMakeLists.txt
```diff
@@ -22,15 +22,20 @@ set(CMAKE_SKIP_RPATH TRUE)
...
 # Setup output folder
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
+enable_testing()
+
 add_subdirectory(${SDK_DIR})
 add_subdirectory(${MP_DIR})
+add_subdirectory(${TESTS_DIR})
```
Mionsz, line 37: I think a flag for enable/disable tests is mandatory. Could you add it?

tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+# Link the executable with gtest and gtest_main
+target_link_libraries(MCM_unit_tests PRIVATE gtest gtest_main ${MP_LIB})
+target_include_directories(MCM_unit_tests PUBLIC
+    ${CMAKE_SOURCE_DIR}/media-proxy/include
+    # Include generated *.pb.h files
+    ${CMAKE_BINARY_DIR}/media-proxy/
+)
+
+# Add tests to CTest
+add_test(NAME MCM_unit_tests COMMAND MCM_unit_tests)
```
Mionsz, line 18: can we have one approach to CAPITAL_letters/CamelCase, :-p?
I think this should be lower_case `mcm_unit_tests` instead of `MCM_unit_tests`.

CMakeLists.txt
```diff
@@ -22,15 +22,20 @@ set(CMAKE_SKIP_RPATH TRUE)
...
+
+target_include_directories(media_proxy_lib PUBLIC ${SDK_INCLUDE_DIR})
+target_link_libraries(media_proxy_lib PUBLIC mcm_dp)
```
Mionsz, line 40: Great catch and idea for renaming component.
What would you say for fitting there also the `sdk` acronym?  Like `media_proxy_sdk_lib` or any other variation so that name will become fully descriptive?

Mionsz (changes requested):
Great work - minor annotation from my side for you to consider and we can have this PR merged.

CMakeLists.txt
Sakoram: My intention was not to rename the component. I created the media_proxy_lib just to be able to link the media_proxy code to the unit tests. Before and after my changes, only the media_proxy executable is built, the media_proxy shared library is not created, so I'm not sure if the product of the compilation process can be called ‘sdk’.

tests/unit/CMakeLists.txt
Sakoram: renamed to mcm_unit_tests

CMakeLists.txt
Sakoram: I've add the flag, but set it to build the test by default.

Mionsz (approved):
Great job, for me this is ready to merge.

PR #176, opened 2024-08-09
Update github_pages_update.yml

Mionsz:
Update github_pages_update.yml
- Fix issue with build and publish script.


soopel (approved):
All looks good to me.

PR #180, opened 2024-08-26
Run basic implementation of AF_XDP/eBPF

Mionsz:
Run basic implementation of AF_XDP/eBPF using mtl native functions and MtlManager. The enablement of `native_af_xdp` for interfaces used in media-proxy workloads required minor impact changes in Dockerfile build sequence as well as adjustments to proxy_context.cc

Biggest issue was encountered for number of queues that are predefined and hardcoded in media-proxy. Current approach treats ax_xdp as alternative case and uses value `62` instead of `128`  for `rx_queues_cnt` and `tx_queues_cnt`.

Test scripts are partially developed and currently not enabled. After confirmed one node deployment possibility I will adjust and enable them.

### DIFF
--- a/media-proxy/README.md
+++ b/media-proxy/README.md
@@ -42,7 +42,7 @@ Notice that the device must have a pre-assigned IP address. The `--ip` parameter
 
 > [!CAUTION]
 > `MtlManager` from the Media-Transport-Library `manager` subdirectory must be running.
-> Only a device physical function with a preconfigured IP address can be used for the `native_af_xdp` mode.
+> Only a device physical function with a pre-configured IP address can be used for the `native_af_xdp` mode.
 
 ### Docker
 The Media Proxy can be run as a docker container.
--- a/media-proxy/README.md
+++ b/media-proxy/README.md
@@ -37,12 +37,12 @@ Usage: media_proxy [OPTION]
 
 #### Run media-proxy using `native_af_xdp`
 
-To use media-proxy with native `af_xdp/ebpf` device name should be provided with prefix `native_af_xdp:`, for example `media-proxy --dev native_af_xdp:ens259f0np0`.
+To use media-proxy with the native `af_xdp/ebpf` device a device name should be provided with the `native_af_xdp:` prefix, for example `media-proxy --dev native_af_xdp:ens259f0np0`.
 Notice that device must have a preassigned IP address and `--ip` parameter will not work.
 
 > [!CAUTION]
-> `MtlManager`, from Media-Transport-Library `manager` subdirectory, must be running
-> Only device physical function with preconfigured IP address can be used for `native_af_xdp` mode.
+> `MtlManager` from the Media-Transport-Library `manager` subdirectory must be running.
+> Only a device physical function with a preconfigured IP address can be used for the `native_af_xdp` mode.
 
 ### Docker
 The Media Proxy can be run as a docker container.
--- a/media-proxy/README.md
+++ b/media-proxy/README.md
@@ -35,6 +35,15 @@ Usage: media_proxy [OPTION]
 -t, --tcp=port_number   Port number for TCP socket controller (defaults: 8002).
 ```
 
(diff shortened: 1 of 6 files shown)


soopel (approved):
Looks good to me.

media-proxy/README.md
```diff
@@ -35,6 +35,15 @@ Usage: media_proxy [OPTION]
 -t, --tcp=port_number   Port number for TCP socket controller (defaults: 8002).
 ```
 
+#### Run media-proxy using `native_af_xdp`
+
+To use media-proxy with native `af_xdp/ebpf` device name should be provided with prefix `native_af_xdp:`, for example `media-proxy --dev native_af_xdp:ens259f0np0`.
+Notice that device must have a preassigned IP address and `--ip` parameter will not work.
+
+> [!CAUTION]
+> `MtlManager`, from Media-Transport-Library `manager` subdirectory, must be running
```
ko80, line 44:
```suggestion
> `MtlManager`, from Media-Transport-Library `manager` subdirectory, must be running.
```

media-proxy/README.md
```diff
@@ -35,6 +35,15 @@ Usage: media_proxy [OPTION]
...
+> Only device physical function with preconfigured IP address can be used for `native_af_xdp` mode.
```
ko80, line 45:
```suggestion
> Only a device physical function with a preconfigured IP address can be used for the `native_af_xdp` mode.
```

ko80 (changes requested):
Please add more details on how to configure devices for AF_XDP/eBPF on a single node with an E810 NIC card and setup all environment for that. Provide an example of configuring Tx and Rx sides when sending data in that mode.

tests/single-node-sample-apps/README.md
```diff
@@ -31,6 +31,7 @@ sudo ./test.sh <test-option> <pf-bdf> <input-file> <duration> <frames_number> <w
    * `memif` – use a memif direct connection
    * `st20` – use an ST20 connection via media proxy
    * `st22` – use an ST22 connection via media proxy
+   * `af_xdp` - use native_af_xdp connection for media proxy. MtlManager required.
```
ko80, line 34:
```suggestion
   * `af_xdp` - use a native_af_xdp connection for media proxy. MtlManager is required to be running.
```

ko80:
Could you elaborate more on why exactly `62` queues are used, not `128`? Lack of resources? Why not `64`?

Mionsz:
@ko80 There are 64 queues hardcoded in the MTL but the first one is reserved for system usage. Still when I set it to 63 it does not work (something in logic may be wrong), the 62 is the first number working

media-proxy/README.md
```diff
@@ -35,6 +35,15 @@ Usage: media_proxy [OPTION]
 -t, --tcp=port_number   Port number for TCP socket controller (defaults: 8002).
 ```
 
+#### Run media-proxy using `native_af_xdp`
+
+To use media-proxy with the native `af_xdp/ebpf` device a device name should be provided with the `native_af_xdp:` prefix, for example `media-proxy --dev native_af_xdp:ens259f0np0`.
+Notice that device must have a preassigned IP address and `--ip` parameter will not work.
```
ko80, line 41:
```suggestion
Notice that the device must have a pre-assigned IP address. The `--ip` parameter is not applied in this mode.
```

media-proxy/README.md
```diff
@@ -35,6 +35,15 @@ Usage: media_proxy [OPTION]
...
+
+> [!CAUTION]
+> `MtlManager` from the Media-Transport-Library `manager` subdirectory must be running.
+> Only a device physical function with a preconfigured IP address can be used for the `native_af_xdp` mode.
```
ko80, line 45:
```suggestion
> Only a device physical function with a pre-configured IP address can be used for the `native_af_xdp` mode.
```

ko80 (approved):
LGTM

PR #183, opened 2024-08-30
Cleanup ffmpeg plugin build scripts

tszumski:
Remove  jpeg xs patch application to ffmpeg because:
1. Jpeg xs is not required as direct ffmpeg-plugin,
2. Jpeg xs is used internally by MTL (kahawai.json), call stack: FFMPEG -> MCM -> MTL -> JPEG-XS


ko80 (approved):
Looks good to me

PR #184, opened 2024-09-02
Extending sample applications - mainly parameters and switches

MateuszGrabuszynski:
The sample applications can be used to test the existing MCM implementation. This effort is to cleanup the current apps and add handling for more ST standards.

Before, the sample app was working only for very limited set of video-related options. At the end, this PR also allows using audio and ancillary options of SMPTE ST2110 standards, that are already implemented in Media-Communications-Mesh.

The code provided should be treated as a milestone (partially finalized version). It requires further cleanup - removing all of to-dos, deleting remaining unnecessary comments, and splitting into more files to ensure better readability.


MateuszGrabuszynski:
The last commit (edc9ed6) contains code, which sends audio data that has similarity with the original, but, at the same time, is severely malformed (contains around 1/8th of the original data). Most probable reason is a wrong way of counting of bytes per ptime (single audio frame).

MateuszGrabuszynski:
The reason of audio being malformed was setting `total_num` to `300` by default (`-n` parameter). While it's `0` (not limited), then it works (almost) fine (cutting the end of the audio file). At the moment, this parameter does not work properly for audio (further investigation is needed to make it work as intended).
Further info to be placed in a comment when changing from **Draft** to **Ready**.

sdk/samples/sample_common.c
```diff
@@ -0,0 +1,218 @@
...
+        case PIX_FMT_YUV422P_10BIT_LE:
+        default:
+            size = pixels * 2 * 2;
+    }
+    if (interlaced) size /= 2; // TODO: Check if all formats support interlace
+    return (int)size;
+}
+
+/* print a description of all supported options */
+void usage(FILE* fp, const char* path, int is_sender)
```
Mionsz, line 133:
```suggestion
static void usage(FILE* fp, const char* path, int is_sender)
```

sdk/samples/sample_common.c
```diff
@@ -0,0 +1,218 @@
...
+        DEFAULT_AUDIO_PTIME);
+    fprintf(fp, "-c, --audio_channels=<channels>\t\t"
+                "Define number of audio channels [1|2] (default: %d)\n",
+        DEFAULT_AUDIO_CHANNELS);
+    fprintf(fp, "-------------------------------------- ANCILLARY (ST4x) --------------------------------------\n");
+    fprintf(fp, "-q, --anc_type=<anc_type>\t\t"
+                "Define anc type [frame|rtp] (default: %s)\n",
+        DEFAULT_ANC_TYPE);
+    fprintf(fp, "\n");
+}
```
Mionsz, line 218:
```suggestion
}

```

Mionsz (commented):
Great work!

tests/single-node-sample-apps/test.sh
```diff
@@ -344,15 +353,172 @@ function run_test_stXX() {
...
+    return $(($error || $timeout || $recver_app_timeout || $sender_app_timeout))
+}
+
 info "Test MCM Tx/Rx for Single Node"
 info "  Binary directory: $(realpath $bin_dir)"
 info "  Output directory: $(realpath $out_dir)"
 info "  Input file path: $input_file"
 info "  Input file size: $(stat -c%s $input_file) byte(s)"
-info "  Frame size: $width x $height, FPS: $fps"
+info "  Frame size: $width x $height, FPS: $fps / Audio ptime: $audio_ptime / Ancilary type: $anc_type"
```
ko80, line 517: No-no, this line mustn't print data for both video and audio.
Rework this to print only video parameters when testing video, OR only audio parameters when testing audio. Same for st40/41

ko80 (changes requested):

ko80:
Great work done @MateuszGrabuszynski!

I just want to get back to the idea of sample apps. The sample apps are for the user to understand how to use MCM in their use cases. It is always better to provide simple examples of usage. In this PR, you're making the sample apps overloaded with implementation of all types of payload. I'm unsure if the end user will have such a use case to work with all types of load within the same app. I think it would be way better to provide separate examples (sample apps) for video, audio, and ancillary use cases.

I have the same impression of adding more stuff to the test. I suggest creating separate test scripts for audio and ancillary instead of making the original one more and more complex.

sdk/src/mcm_dp.c
```diff
@@ -267,6 +267,15 @@ mcm_conn_context* mcm_create_connection(mcm_conn_param* param)
...
         break;
+    case PROTO_TCP:
+        log_warn("Unsupported protocol: TCP (%d)", param->protocol);
+        break;
+    case PROTO_HTTP:
+        log_warn("Unsupported protocol: HTTP (%d)", param->protocol);
+        break;
+    case PROTO_GRPC:
+        log_warn("Unsupported protocol: GRPC (%d)", param->protocol);
+        break;
```
ko80, line 278: Could you elaborate more on why we need these cases instead of the existing default printing?

MateuszGrabuszynski: Existing default printing is theoretically enough, but not very descriptive, e.g. `WARN - Unsupported protocol: 3` does not say much about which protocol is supposedly unsupported. I know it may sound like one step too far, but it may shorten the debugging time.

ko80: Mmm, it's redundant, really.

tests/single-node-sample-apps/test.sh
MateuszGrabuszynski: Good point - I tried to fix that in such way (03e024e):

## ST20 (Raw video)
```shell
sudo LD_LIBRARY_PATH=$(pwd)/../../out/lib ./test.sh st20 0000:c0:00.1 ../../1280x720_yuv422p10le_60fps.yuv 1 60 1280 720 yuv422p10le
```
```text
2024-09-16 14:56:59 [INFO] Test MCM Tx/Rx for Single Node
2024-09-16 14:56:59 [INFO]   Binary directory: /home/gta/MCM/out/bin
2024-09-16 14:56:59 [INFO]   Output directory: /home/gta/MCM/tests/single-node-sample-apps/out
2024-09-16 14:56:59 [INFO]   Test option: st20
2024-09-16 14:56:59 [INFO]   NIC PF: 0000:c0:00.1
2024-09-16 14:56:59 [INFO]   Input file path: /home/gta/MCM/1280x720_yuv422p10le_60fps.yuv
2024-09-16 14:56:59 [INFO]   Input file size: 442368000 byte(s)
2024-09-16 14:56:59 [INFO]   VIDEO OPTIONS:
2024-09-16 14:56:59 [INFO]     Duration: 1s
2024-09-16 14:56:59 [INFO]     Number of frames: 60
2024-09-16 14:56:59 [INFO]     Frame size: 1280 x 720
2024-09-16 14:56:59 [INFO]     FPS: yuv422p10le
2024-09-16 14:56:59 [INFO]     Pixel format: yuv422p10le
...
```

## ST22 (Compressed video)
```shell
sudo LD_LIBRARY_PATH=$(pwd)/../../out/lib ./test.sh st22 0000:c0:00.1 ../../1280x720_yuv422p10le_60fps.jxs 1 60 1280 720 yuv422p10le
```
```text
2024-09-16 14:57:40 [INFO] Test MCM Tx/Rx for Single Node
2024-09-16 14:57:40 [INFO]   Binary directory: /home/gta/MCM/out/bin
2024-09-16 14:57:40 [INFO]   Output directory: /home/gta/MCM/tests/single-node-sample-apps/out
2024-09-16 14:57:40 [INFO]   Test option: st22
2024-09-16 14:57:40 [INFO]   NIC PF: 0000:c0:00.1
2024-09-16 14:57:40 [INFO]   Input file path: /home/gta/MCM/1280x720_yuv422p10le_60fps.jxs
stat: cannot statx '/home/gta/MCM/1280x720_yuv422p10le_60fps.jxs': No such file or directory
2024-09-16 14:57:40 [INFO]   Input file size:  byte(s)
2024-09-16 14:57:40 [INFO]   VIDEO OPTIONS:
2024-09-16 14:57:40 [INFO]     Duration: 1s
2024-09-16 14:57:40 [INFO]     Number of frames: 60
2024-09-16 14:57:40 [INFO]     Frame size: 1280 x 720
2024-09-16 14:57:40 [INFO]     FPS: yuv422p10le
2024-09-16 14:57:40 [INFO]     Pixel format: yuv422p10le
```

## ST30 (Audio)
```shell
sudo LD_LIBRARY_PATH=$(pwd)/../../out/lib ./test.sh st30 0000:c0:00.1 ../../c0_44100_pcm16.wav pcm16 44k 1ms 1
```
```text
2024-09-16 14:53:53 [INFO] Test MCM Tx/Rx for Single Node
2024-09-16 14:53:53 [INFO]   Binary directory: /home/gta/MCM/out/bin
2024-09-16 14:53:53 [INFO]   Output directory: /home/gta/MCM/tests/single-node-sample-apps/out
2024-09-16 14:53:53 [INFO]   Test option: st30
2024-09-16 14:53:53 [INFO]   NIC PF: 0000:c0:00.1
2024-09-16 14:53:53 [INFO]   Input file path: /home/gta/MCM/c0_44100_pcm16.wav
2024-09-16 14:53:53 [INFO]   Input file size: 88278 byte(s)
2024-09-16 14:53:53 [INFO]   AUDIO OPTIONS:
2024-09-16 14:53:53 [INFO]     Audio type: pcm16
2024-09-16 14:53:53 [INFO]     Audio sampling: 44k
2024-09-16 14:53:53 [INFO]     Audio ptime: 1ms
2024-09-16 14:53:53 [INFO]     Audio channels: 1
...
```

## ST40 (Ancillary)
```shell
sudo LD_LIBRARY_PATH=$(pwd)/../../out/lib ./test.sh st40 0000:c0:00.1 ../../CCtest.srt frame
```

```text
2024-09-16 15:17:26 [INFO] Test MCM Tx/Rx for Single Node
2024-09-16 15:17:26 [INFO]   Binary directory: /home/gta/MCM/out/bin
2024-09-16 15:17:26 [INFO]   Output directory: /home/gta/MCM/tests/single-node-sample-apps/out
2024-09-16 15:17:26 [INFO]   Test option: st40
2024-09-16 15:17:26 [INFO]   NIC PF: 0000:c0:00.1
2024-09-16 15:17:26 [INFO]   Input file path: /home/gta/MCM/CCtest.srt
2024-09-16 15:17:26 [INFO]   Input file size: 38214 byte(s)
2024-09-16 15:17:26 [INFO]   ANCILLARY OPTIONS:
2024-09-16 15:17:26 [INFO]     Ancillary type: frame
```

sdk/src/mcm_dp.c
MateuszGrabuszynski: I'll be honest, I was quite confused when I encountered that for the first time. Is there any standard that numbers these protocols in the same manner?

If we really want to get back to the previous solution, what if we skip the number, leaving `WARN - Unsupported protocol`, or at least put it in the bracket `WARN - Unsupported protocol (3)`?

ko80: I vote for `WARN - Unsupported protocol (3)`

MateuszGrabuszynski:
> I think it would be way better to provide separate examples (sample apps) for video, audio, and ancillary use cases.
> 
> I have the same impression of adding more stuff to the test. I suggest creating separate test scripts for audio and ancillary instead of making the original one more and more complex.

I totally understand and thus agree, but considering it would require a set of tests to be redone - I want this PR to be a temporary working milestone, so we don't miss the deadline. And after merging, I can work on the split. Baby steps :)

ko80:
> > I think it would be way better to provide separate examples (sample apps) for video, audio, and ancillary use cases.
> > I have the same impression of adding more stuff to the test. I suggest creating separate test scripts for audio and ancillary instead of making the original one more and more complex.
> 
> I totally understand and thus agree, but considering it would require a set of tests to be redone - I want this PR to be a temporary working milestone, so we don't miss the deadline. And after merging, I can work on the split. Baby steps :)

Agree. So please add a very visible comment in the this PR and in the code of the test script that the test script must be split into simpler test scripts in the future.

sdk/src/mcm_dp.c
MateuszGrabuszynski: > I vote for `WARN - Unsupported protocol (3)`

Done in 61e3366.

sdk/samples/sample_common.c
```diff
@@ -0,0 +1,218 @@
...
+        case PIX_FMT_YUV422P_10BIT_LE:
+        default:
+            size = pixels * 2 * 2;
+    }
+    if (interlaced) size /= 2; // TODO: Check if all formats support interlace
+    return (int)size;
+}
+
+/* print a description of all supported options */
+static void usage(FILE* fp, const char* path, int is_sender)
```
ko80, line 133: I don't agree with extracting this from sender_app and recver_app and putting here with the `is_sender` argument to differentiate the output according to where it was called: from sender_app or recver_app. This is overengineering and premature optimization at the same time.

Please move the `usage()` implementation back to sender_app and recver_app respectively.

sdk/samples/sender_app.c
```diff
@@ -185,20 +106,28 @@ int main(int argc, char** argv)
...
+        { "file", required_argument, NULL, 'b' },
         { "pix_fmt", required_argument, NULL, 'x' },
+        { "audio_type", required_argument, NULL, 'a' },
+        { "audio_format", required_argument, NULL, 'j' },
+        { "audio_sampling", required_argument, NULL, 'g' },
+        { "audio_ptime", required_argument, NULL, 'e' },
+        { "audio_channels", required_argument, NULL, 'c' },
+        { "anc_type", required_argument, NULL, 'q' },
+        { "number", required_argument, NULL, 'n' },
+        { "loop", required_argument, NULL, 'l' },
```
ko80, line 122: Adding arguments here? Please reflect the change in README.md for the user to be familiar with them.

sdk/samples/sender_app.c
```diff
@@ -432,7 +440,8 @@ int main(int argc, char** argv)
 
         frame_count++;
 
-        if (frame_count >= total_num) {
+        if (param.payload_type != PAYLOAD_TYPE_ST30_AUDIO \
+            && total_num > 0 && frame_count >= total_num) {
```
ko80, line 444:
```suggestion
        if (param.payload_type != PAYLOAD_TYPE_ST30_AUDIO
            && total_num > 0 && frame_count >= total_num) {
```

sdk/samples/recver_app.c
```diff
@@ -276,38 +221,100 @@ int main(int argc, char** argv)
...
+        frm_size = getFrameSize(pix_fmt, width, height, false);
         break;
     }
 
     strlcpy(param.remote_addr.ip, recv_addr, sizeof(param.remote_addr.ip));
     strlcpy(param.local_addr.port, recv_port, sizeof(param.local_addr.port));
     strlcpy(param.remote_addr.port, send_port, sizeof(param.remote_addr.port));
     strlcpy(param.local_addr.ip, send_addr, sizeof(param.local_addr.ip));
+    fprintf(stdout, "LOCAL: %s:%s\n", param.local_addr.ip, param.local_addr.port);
+    fprintf(stdout, "REMOTE: %s:%s\n", param.remote_addr.ip, param.remote_addr.port);
```
ko80, line 317:
```suggestion
    printf("LOCAL: %s:%s\n", param.local_addr.ip, param.local_addr.port);
    printf("REMOTE: %s:%s\n", param.remote_addr.ip, param.remote_addr.port);
```

sdk/samples/sender_app.c
```diff
@@ -316,37 +262,100 @@ int main(int argc, char** argv)
...
         break;
     }
 
     strlcpy(param.remote_addr.ip, send_addr, sizeof(param.remote_addr.ip));
     strlcpy(param.remote_addr.port, send_port, sizeof(param.remote_addr.port));
     strlcpy(param.local_addr.ip, send_addr, sizeof(param.local_addr.ip));
     strlcpy(param.local_addr.port, send_port, sizeof(param.local_addr.port));
-    frame_size = getFrameSize(pix_fmt, width, height, false);
+    fprintf(stdout, "LOCAL: %s:%s\n", param.local_addr.ip, param.local_addr.port);
+    fprintf(stdout, "REMOTE: %s:%s\n", param.remote_addr.ip, param.remote_addr.port);
```
ko80, line 358:
```suggestion
    printf("LOCAL: %s:%s\n", param.local_addr.ip, param.local_addr.port);
    printf("REMOTE: %s:%s\n", param.remote_addr.ip, param.remote_addr.port);
```

sdk/samples/sender_app.c
```diff
@@ -168,10 +81,18 @@ int main(int argc, char** argv)
...
+    char audio_ptime[6] = DEFAULT_AUDIO_PTIME;
+    char anc_type[5] = DEFAULT_ANC_TYPE;
+    char payload_codec[6] = DEFAULT_PAYLOAD_CODEC;
+    uint32_t audio_channels = DEFAULT_AUDIO_CHANNELS;
+
     mcm_conn_context* dp_ctx = NULL;
     mcm_conn_param param = { 0 };
     mcm_buffer* buf = NULL;
-    uint32_t total_num = 300;
+    uint32_t total_num = DEFAULT_TOTAL_NUM;
```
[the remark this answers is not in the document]
MateuszGrabuszynski: There's no default defined in any readme. This was changed on purpose. 0 = do not limit, send until the end of file / stopped by the user. It has to be 0 for audio, otherwise the output is mumbled. We may change the code to mitigate that later.

From usage/README:
```text
-n, --number=<number_of_frames>         Total frame number to send (default = inf: 0)
```

sdk/samples/sender_app.c
MateuszGrabuszynski: Added to README.md (with accordance to usage/help)

sdk/samples/recver_app.c
MateuszGrabuszynski: Fixed

sdk/samples/sender_app.c
MateuszGrabuszynski: Fixed

sdk/samples/sample_common.c
MateuszGrabuszynski: Sorry, but I politely disagree. This function is used just for help message with usage, and does nothing else. This message has 20 switches for both apps, 3 for just sender, 1 for just receiver (with a comment in line 174 for clarity). So differing by just a few switches, 4/24 ~= 16%, to be exact.

In my opinion, duplicating that in 2 files will open the doors to misalign the switches between sender and receiver, and double the work when extending/changing/removing anything.

If your suggestion is to change the name of the variable, please provide a proposed name.

sdk/README.md
```diff
@@ -42,12 +42,33 @@ Sample code for the application which send out data to others.
...
+-i, --rcv_port=<port_number>            Receiver's port number (default: 9001)
+-s, --send_ip=<ip_address>              Sender's IP address (default: 127.0.0.1)
+-p, --send_port=<port_number>           Sender's port number (default: 9001)
+-o, --protocol=<protocol_type>          Set protocol type (default: auto)
+-t, --type=<payload_type>               Payload type (default: st20)
+-k, --socketpath=<socket_path>          Set memif socket path (default: /run/mcm/mcm_rx_memif.sock)
+-m, --master=<is_master>                Set memif conn is master (default: 1 for sender, 0 for recver)
+-d, --interfaceid=<interface_id>        Set memif conn interface id (default: 0)
+-b, --file=<input_file>                 Input file name (optional)
+-l, --loop=<is_loop>                    Set infinity loop sending (default: 0)
```
ko80, line 57:
```suggestion
-l, --loop=<is_loop>                    Set infinite loop sending (default: 0)
```

sdk/README.md
```diff
@@ -42,12 +42,33 @@ Sample code for the application which send out data to others.
...
+--------------------------------------   VIDEO (ST2x)   --------------------------------------
+-w, --width=<frame_width>               Width of test video frame (default: 1920)
+-h, --height=<frame_height>             Height of test video frame (default: 1080)
+-f, --fps=<video_fps>                   Test video FPS (frame per second) (default: 30.00)
+-x, --pix_fmt=<pixel_format>            Pixel format (default: yuv422p10le)
+-n, --number=<number_of_frames>         Total frame number to send (default = inf: 0)
```
ko80, line 63:
```suggestion
-n, --number=<number_of_frames>         Total frame number to send (default: 0, i.e. infinite)
```

sdk/README.md
```diff
@@ -57,10 +78,31 @@ Sample code for the application which receive data from others.
...
+-d, --interfaceid=<interface_id>        Set memif conn interface id (default: 0)
+-b, --dumpfile=<file_name>              Save stream to local file (example: data-sdk.264)
+--------------------------------------   VIDEO (ST2x)   --------------------------------------
+-w, --width=<frame_width>               Width of test video frame (default: 1920)
+-h, --height=<frame_height>             Height of test video frame (default: 1080)
+-f, --fps=<video_fps>                   Test video FPS (frame per second) (default: 30.00)
+-x, --pix_fmt=<pixel_format>            Pixel format (default: yuv422p10le)
+--------------------------------------   AUDIO (ST3x)   --------------------------------------
+-a, --audio_type=<audio_type>           Define audio type [frame|rtp] (default: frame)
+-j, --audio_format=<audio_format>       Define audio format [pcm8|pcm16|pcm24|am824] (default: pcm16)
```
ko80, line 100: I guess, pcm24 is supposed to be default

sdk/samples/sample_common.c
ko80: Mateusz, I'm not insisting on cancelling that change. You have your strong arguments to keep that this way – this is fine, I agree.

ko80: @Mionsz What do you think?

sdk/README.md
MateuszGrabuszynski: It is to emphasize that default value is 0, which means infinite.

ko80: Not sure if the user understands what `inf` is. I suggest changing to `default: 0, i.e. infinite`

sdk/README.md
MateuszGrabuszynski: Good catch!
```c
#define DEFAULT_AUDIO_FORMAT "pcm16"
```

sdk/README.md
```diff
@@ -57,10 +78,31 @@ Sample code for the application which receive data from others.
...
+-g, --audio_sampling=<audio_sampling>   Define audio sampling [48k|96k|44k] (default: 48k)
+-e, --audio_ptime=<audio_ptime>         Define audio ptime [1ms|125us|250us|333us|4ms|80us|1.09ms|0.14ms|0.09ms] (default: 1ms)
+-c, --audio_channels=<channels>         Define number of audio channels [1|2] (default: 1)
```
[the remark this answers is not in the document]
MateuszGrabuszynski: Here, it seems the default is set to 1.
```c
#define DEFAULT_AUDIO_CHANNELS 1
```

I have to re-check all of those.

sdk/README.md
MateuszGrabuszynski: Don't get me wrong, "id est" fits here the best, but I'd rather stick to `default = infinite: 0`. I simply know "i.e." is too often wrongly understood to be equivalent of "e.g.".

sdk/README.md
MateuszGrabuszynski: Quickly double-checked on the README. As the whole section was copied directly from runs of `./sender_app -h` and `./recver_app -h` where those values are printed from the defaults' definitions, all of them are fine.

sdk/README.md
ko80: Agree

MateuszGrabuszynski:
Decided, that I will re-check if it runs properly before merging.

MateuszGrabuszynski:
Reverting to **Draft** state, so we don't accidentally merge it now.

The audio portion does work ONLY for lengths that are multipliers of ptime, which is not true for real-case scenarios. As we do not want to confuse potential users with invalid sample applications, the solution needs to refined for the next release, to cover all possible audio lengths.

PR #185, opened 2024-09-06
Fix audio buffer size + Rework of PCM16/24 support

ko80:
* Fix audio buffer size in memif protocol mode
* Add compatibility check for audio parameters in FFmpeg plugin
* Update documentation on audio ptime options
* Rework MCM audio FFmpeg plugin to support PCM 16/24

### DIFF
--- a/sdk/src/mcm_dp.c
+++ b/sdk/src/mcm_dp.c
@@ -6,6 +6,7 @@
 #include <bsd/string.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <errno.h>
 
 #include "memif_impl.h"
 #include "udp_impl.h"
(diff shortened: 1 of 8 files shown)


sdk/src/mcm_dp.c
```diff
@@ -6,17 +6,120 @@
...
 
+/* Calculate audio buffer size based on ST2110-30 related parameters */
+static int mcm_calc_audio_buffer_size(mcm_audio_args *params, uint32_t *size)
+{
+    uint32_t sample_size = 0;
+    uint32_t sample_num = 0;
+
+    switch (params->format) {
+    case AUDIO_FMT_PCM8:
+        sample_size = 1;
```
soopel, line 25: Is there any way to automate this formula somehow? Dependency between format and sample_size value?

sdk/src/mcm_dp.c
```diff
@@ -6,17 +6,120 @@
...
+        break;
+    case AUDIO_FMT_AM824:
+        sample_size = 4;
+        break;
+    default:
+        *size = 0;
+        return -EINVAL;
+    }
+
+    switch (params->sampling) {
```
soopel, line 41: Same here. Is there any way to automate this formula somehow? Dependency between sampling and sample_num value?

sdk/src/mcm_dp.c
```diff
@@ -6,17 +6,120 @@
...
+            break;
+        case AUDIO_PTIME_80US:
+            sample_num = 4;
+            break;
+        default:
+            *size = 0;
+            return -EINVAL;
+        }
+        break;
+    case AUDIO_SAMPLING_96K:
```
soopel, line 67: Same as above

sdk/src/mcm_dp.c
MateuszGrabuszynski: https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/dde0aaea3cb2d26ac5cac135133bda303814c818/sdk/include/mcm_dp.h#L112-L118
Perhaps:
```c
sample_size = params->format + 1;
```

ko80: > Is there any way to automate this formula somehow? Dependency between format and sample_size value?

This function is implemented the same way as in MTL. Take a look https://github.com/OpenVisualCloud/Media-Transport-Library/blob/3127c88b8a73d59520189c72254559a436125ed3/lib/src/st2110/st_fmt.c#L1037

I suggest to leave it as it is in order to stay aligned with MTL implementation.

ko80: > https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/dde0aaea3cb2d26ac5cac135133bda303814c818/sdk/include/mcm_dp.h#L112-L118
> 
> 
> Perhaps:
> ```c
> sample_size = params->format + 1;
> ```

I can't agree with that. You're suggesting to remove clarity from the code. A golden rule here is "Do not optimize prematurely".

sdk/src/mcm_dp.c
ko80: See my comment above https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/185#discussion_r1751650182

sdk/src/mcm_dp.c
```diff
@@ -6,17 +6,120 @@
...
+            break;
+        default:
+            *size = 0;
+            return -EINVAL;
+        }
+        break;
+    default:
+        *size = 0;
+        return -EINVAL;
+    }
```
MateuszGrabuszynski, line 111:
```suggestion
    switch (params->sampling) {
    case AUDIO_SAMPLING_48K:
    case AUDIO_SAMPLING_96K:
        switch (params->ptime) {
        case AUDIO_PTIME_250US:
            sample_num = 12;
            break;
        case AUDIO_PTIME_333US:
            sample_num = 16;
            break;
        case AUDIO_PTIME_4MS:
            sample_num = 192;
            break;
        default:
            *size = 0;
            return -EINVAL;
        }
    case AUDIO_SAMPLING_44K:
        switch (params->ptime) {
        case AUDIO_PTIME_1_09MS:
        case AUDIO_PTIME_1MS:
            sample_num = 48;
            break;
        case AUDIO_PTIME_0_14MS:
        case AUDIO_PTIME_125US:
            sample_num = 6;
            break;
        case AUDIO_PTIME_0_09MS:
        case AUDIO_PTIME_80US:
            sample_num = 4;
            break;
        default:
            *size = 0;
            return -EINVAL;
        }
        break;
    default:
        *size = 0;
        return -EINVAL;
    }
    if (params->sampling == AUDIO_SAMPLING_96K){ sample_num *= 2; }
```

ko80: Do. Not. Optimize. Prematurely.

Mionsz (commented):
After current conversations resolved I see no blocker to merge it.

soopel (approved):
Agree with responses to feedback.

PR #186, opened 2024-09-11
Fix audio buffer size in FFmpeg MCM plugin + fix Rx ring size

ko80:

* Fix audio buffer size in FFmpeg MCM plugin + fix Rx ring size

### DIFF
--- a/ffmpeg-plugin/mcm_audio_tx.c
+++ b/ffmpeg-plugin/mcm_audio_tx.c
@@ -11,6 +11,7 @@
 #include "libavformat/mux.h"
 #include "libavdevice/mcm_common.h"
 #include <mcm_dp.h>
+#include <unistd.h>
 
 typedef struct McmAudioMuxerContext {
     const AVClass *class; /**< Class for private options. */
@@ -28,6 +29,8 @@ typedef struct McmAudioMuxerContext {
     char* ptime;
 
     mcm_conn_context *tx_handle;
+    mcm_buffer *unsent_mcm_buf;
+    int unsent_len;
 } McmAudioMuxerContext;
 
 static int mcm_audio_write_header(AVFormatContext* avctx)
@@ -114,29 +117,50 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
 static int mcm_audio_write_packet(AVFormatContext* avctx, AVPacket* pkt)
 {
     McmAudioMuxerContext *s = avctx->priv_data;
+    size_t frame_size = s->tx_handle->frame_size;
     const uint8_t *data = pkt->data;
-    mcm_buffer *buf = NULL;
     int size = pkt->size;
     int len, err = 0;
 
     while (size > 0) {
-        buf = mcm_dequeue_buffer(s->tx_handle, -1, &err);
-        if (buf == NULL) {
-            av_log(avctx, AV_LOG_ERROR, "Dequeue buffer error %d\n", err);
-            return AVERROR(EIO);
+        char *dest;
+
+        if (!s->unsent_mcm_buf) {
+            s->unsent_mcm_buf = mcm_dequeue_buffer(s->tx_handle, -1, &err);
+            if (s->unsent_mcm_buf == NULL) {
+                av_log(avctx, AV_LOG_ERROR, "Initial dequeue buffer error %d\n", err);
+                return AVERROR(EIO);
+            }
         }
 
-        len = FFMIN(buf->len, size);
-        memcpy(buf->data, data, len);
+        len = FFMIN(frame_size, size + s->unsent_len);
+
+        if (len < frame_size) {
+            memcpy((char *)s->unsent_mcm_buf->data + s->unsent_len, data, size);
+            s->unsent_len += size;
+            break;
+        }
+
+        dest = s->unsent_mcm_buf->data;
+
+        if (s->unsent_len) {
+            memcpy(dest, s->unsent_mcm_buf->data, s->unsent_len);
+            dest += s->unsent_len;
+            len -= s->unsent_len;
+            s->unsent_len = 0;
+        }
+
+        memcpy(dest, data, len);
         data += len;
         size -= len;
 
-        buf->len = len;
-
-        if ((err = mcm_enqueue_buffer(s->tx_handle, buf)) != 0) {
+        err = mcm_enqueue_buffer(s->tx_handle, s->unsent_mcm_buf);
+        if (err) {
             av_log(avctx, AV_LOG_ERROR, "Enqueue buffer error %d\n", err);
             return AVERROR(EIO);
         }
+
+        s->unsent_mcm_buf = NULL;
     }
 
     return 0;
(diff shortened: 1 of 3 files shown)


ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -114,29 +117,50 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
...
+        if (len < frame_size) {
+            memcpy((char *)s->unsent_mcm_buf->data + s->unsent_len, data, size);
+            s->unsent_len += size;
+            break;
+        }
+
+        dest = s->unsent_mcm_buf->data;
+
+        if (s->unsent_len) {
+            memcpy(dest, s->unsent_mcm_buf->data, s->unsent_len);
```
Mionsz, line 147: This is literally `memcpy(dest, dest, s->unsent_len);` so undefined behavior of memcpy for overlapping memory slices.

ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -114,29 +117,50 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
...
 
-        len = FFMIN(buf->len, size);
-        memcpy(buf->data, data, len);
+        len = FFMIN(frame_size, size + s->unsent_len);
+
+        if (len < frame_size) {
+            memcpy((char *)s->unsent_mcm_buf->data + s->unsent_len, data, size);
+            s->unsent_len += size;
+            break;
+        }
```
Mionsz, line 142: Having `if(){}  else{}` instead of `break;` and `static int mcm_audio_write_trailer(AVFormatContext* avctx)` added part wouldn't be better?

ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -114,29 +117,50 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
 static int mcm_audio_write_packet(AVFormatContext* avctx, AVPacket* pkt)
 {
     McmAudioMuxerContext *s = avctx->priv_data;
+    size_t frame_size = s->tx_handle->frame_size;
```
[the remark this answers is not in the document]
ko80: No, they can't. It is guaranteed by design. Both fields are filled in `mcm_audio_write_header()` which is called once before `mcm_audio_write_packet()`.

ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -114,29 +117,50 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
...
 
     while (size > 0) {
-        buf = mcm_dequeue_buffer(s->tx_handle, -1, &err);
-        if (buf == NULL) {
-            av_log(avctx, AV_LOG_ERROR, "Dequeue buffer error %d\n", err);
-            return AVERROR(EIO);
+        char *dest;
+
+        if (!s->unsent_mcm_buf) {
+            s->unsent_mcm_buf = mcm_dequeue_buffer(s->tx_handle, -1, &err);
```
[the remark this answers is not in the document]
ko80: I agree it looks suspicious. But there is no major problem. Let me explain that. We use this memif connection to communicate with Media Proxy only, which controls the connection. If the connection is closed, `mcm_dequeue_buffer()` returns. The flow stays blocked here until the memif library provides an empty buffer to fill. And this is totally fine because if the call is timed and a timeout occurs, the flow will have to break transmission and that's not what we want to happen.

Let's keep this approach as it is for now since it is working.

ffmpeg-plugin/mcm_audio_tx.c
ko80: Not sure I understand that. Could you elaborate more on your concern?

ffmpeg-plugin/mcm_audio_tx.c
ko80: OMG. Great catch Milosz! Let me rework that.

ffmpeg-plugin/mcm_audio_tx.c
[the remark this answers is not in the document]
Mionsz: It is more like a question if you have multiple audio buffers won't this block any other possible operations ongoing (so how the multithreading is handled). Also getting buffer if the queue is full will not work as intended, and should we do an audio frame skip then or wait?

ffmpeg-plugin/mcm_audio_tx.c
ko80: This memcpy has to be removed. It left after refactoring but didn't affect the algorithm. That's why I missed it. I'm going to fix it in a separate commit.

ffmpeg-plugin/mcm_audio_tx.c
Mionsz: I would move the part:
```c
static int mcm_audio_write_trailer(AVFormatContext* avctx)
{
    McmAudioMuxerContext *s = avctx->priv_data;

    if (s->unsent_mcm_buf) {
        int err;

        /* zero the unused rest of the buffer */
        memset((char *)s->unsent_mcm_buf->data + s->unsent_len, 0,
               s->unsent_mcm_buf->len - s->unsent_len);

        err = mcm_enqueue_buffer(s->tx_handle, s->unsent_mcm_buf);
        if (err) {
            av_log(avctx, AV_LOG_ERROR, "Enqueue buffer error %d\n", err);
            return AVERROR(EIO);
        }
    }
```
back here:
```c
        if (len < frame_size) {
            memcpy((char *)s->unsent_mcm_buf->data + s->unsent_len, data, size);
            s->unsent_len += size;
            // break;
            // <---- HERE
        }
```
and have rest of code wrapped in `else { }` statement.
This way you will not exit this method before putting the buffer back.

ffmpeg-plugin/mcm_audio_tx.c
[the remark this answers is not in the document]
ko80: For sure we can't skip anything. That's why infinite waiting is meaningful here. If a skip happens, the queue capacity has to be increased.

When it comes to multiple audio streams, every stream is processed by FFmpeg in a separate thread. And every connection to MCM creates a separate memif queue which is also processed in separate threads. So, the implementation is thread-safe. We don't use any shared data here between the audio streams.

ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -114,29 +117,49 @@ static int mcm_audio_write_header(AVFormatContext* avctx)
...
+
+        dest = s->unsent_mcm_buf->data;
+
+        if (s->unsent_len) {
+            dest += s->unsent_len;
+            len -= s->unsent_len;
+            s->unsent_len = 0;
+        }
+
+        memcpy(dest, data, len);
```
Mionsz, line 152: This will overflow as here is the `if (len >= frame_size)` approach.

ko80: Nope. Here we have `len == frame_size`. It can't be bigger because it is defined this way: `len = FFMIN(frame_size, size + s->unsent_len)`.

So, this copying is safe by design.

Mionsz:
What about 
image
and
image

ko80:
Replaced `memcpy` with `mtl_memcpy` in 3782c527ac4f75bd220a1c1f610ebabde341f292

ko80:
We can't use MTL API functions directly because MCM SDK is not dependent on MCM and isolates the user from the implementation under the hood of MCM. We can schedule some refactoring to improve performance in the future and to achieve that we will have to create MCM SDK API functions to wrap those of MTL.

ffmpeg-plugin/mcm_audio_tx.c
Mionsz: OK, true, now I see it.
`len = FFMIN(frame_size, size + s->unsent_len);`

PR #189, opened 2024-09-16
Format code style

Sakoram:
Use clang-format to format whole project to a style similar to linux kernel code style.

For now I used: 
```
IncludeBlocks: Preserve
SortIncludes: false
```
because reordering includes brakes the build process.


Sakoram:
This PR is outdated. Most of the code in which it improves code style is already after refactoring

PR #192, opened 2024-09-17
Build dependencies and linker fixes

Mionsz:
[Build dependencies and linker fixes](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/192/commits/7341bdc6d396ea4157e52cb18944a5c55f944fe4)
Build dependencies and linker fixes focus is on:
- building all dependencies for mcm as a pre-build phase
- address linker issues that required passing specific paths for libraries to be found under debian
- lots of minor changes is between.

[Test.sh modified and minor fixes to mtl.c and CMakeLists.txt](https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/192/commits/df30c486674b02c650e41bf88625eff43da3eeee)
- Test.sh modified to copy results on error.
- Fixes issue with `segfault` on TxSide
- Minor fixes to `mtl.c` to all `shm_deinit`.
- Minor fixes to CMakeLists.txt.
- Fixes to LogLevel so that it can be controlled.
- General scripting fixes and improvements.

Calling pthread_join after pthread_cancel can result in `err == ESRCH` as pthread can exit faster then we call `pthread_join` - resulting in error thread do not exist. We should treat this as a non-error.


sdk/CMakeLists.txt
```diff
@@ -50,15 +50,15 @@ set(CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_PREFIX}/lib)
 set(CMAKE_INSTALL_IPATH ${CMAKE_INSTALL_PREFIX}/include)
 
 aux_source_directory(${MCMDP_SRC_DIR} sdk_srcs)
-add_library(mcm_dp SHARED ${sdk_srcs})
```
Sakoram, line 53: Why this change? I though that the idea is to create a dynamic library and .so file of SDK which a user can load. The header files, SDK .so, and already compiled media_proxy is all what they need. If I'm not wrong after the change from SHARED to OBJECT the .so wont be created, so a user can't easily link their code with the SDK.

Sakoram: ok, now I see it. You then create a shared lib with: `add_library(mcm_dp_shared SHARED $<TARGET_OBJECTS:mcm_dp>)`. Never mind.

PR #195, opened 2024-09-19
Fix how memif_event_thread is canceled

Sakoram:
Resources may not have been deallocated at all or
deallocated while the thread was still running


ko80 (approved):
LGTM

PR #196, opened 2024-09-20
Fix issues in RDMA implementation + improve memif buffer allocation handling

Sakoram:
- Remove unnecessary pacing from sender_app
- Add timeout for memif_buffer_alloc()
- Increase ring size for RDMA
- Improve test-rdma.sh

### DIFF
--- a/media-proxy/src/shm_memif_common.c
+++ b/media-proxy/src/shm_memif_common.c
@@ -199,19 +199,21 @@ int tx_on_disconnect(memif_conn_handle_t conn, void* priv_data)
 
 int memif_buffer_alloc_timeout(memif_conn_handle_t conn, uint16_t qid,
                                memif_buffer_t * bufs, uint16_t count, uint16_t * count_out,
-                               uint32_t size, unsigned int timeout_ms) {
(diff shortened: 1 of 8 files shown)


media-proxy/src/shm_memif_common.c
```diff
@@ -196,3 +196,22 @@ int tx_on_disconnect(memif_conn_handle_t conn, void* priv_data)
...
+
+    while (waited_half_ms < timeout_ms * 2) {
+        err = memif_buffer_alloc(conn, qid, bufs, count, count_out, size);
+        if (err == MEMIF_ERR_SUCCESS) {
+            return err;
+        } else if (err != MEMIF_ERR_NOBUF_RING) {
+            return err;
+        }
+        usleep(500);
+        waited_half_ms += 1;
```
ko80, line 214: I suppose the below would be easier to understand
```suggestion
        if (err == MEMIF_ERR_NOBUF_RING) {
            usleep(500);
            waited_half_ms += 1;
            continue;
        }
        if (err)
            return err;
        else
            return 0;
```

ko80 (approved):
Looks good

PR #201, opened 2024-09-23
MCM st22 fix to media_proxy segfault

Mionsz:
MCM st22 fix to media_proxy segfault:

- added fix for media_proxy exit with segfault

### DIFF
--- a/media-proxy/src/mtl.c
+++ b/media-proxy/src/mtl.c
@@ -1164,12 +1164,12 @@ int tx_st22p_shm_deinit(tx_st22p_session_context_t* tx_ctx)
 
     err = pthread_cancel(tx_ctx->memif_event_thread);
     if (err) {
-        ERROR("%s: Error canceling thread: %s", __func__, strerror(err));
+        ERROR("%s: Error canceling thread: (%d) %s", __func__, err, strerror(err));
     }
 
     err = pthread_join(tx_ctx->memif_event_thread, NULL);
     if (err && err != ESRCH) {
-        ERROR("%s: Error joining thread: %s", __func__, strerror(err));
+        ERROR("%s: Error joining thread: (%d) %s", __func__, err, strerror(err));
     }
 
     /* free-up resources */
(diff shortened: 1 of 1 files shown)


media-proxy/src/mtl.c
```diff
@@ -1162,14 +1162,16 @@ int tx_st22p_shm_deinit(tx_st22p_session_context_t* tx_ctx)
         return -1;
     }
 
-    err = pthread_cancel(tx_ctx->memif_event_thread);
-    if (err) {
-        ERROR("%s: Error canceling thread: %s", __func__, strerror(err));
-    }
+    if(!tx_ctx->stop) {
```
Sakoram, line 1165: I really don't like how `tx_ctx->stop` is used here.

The purpose of it is to signal to `rx_st22p_frame_thread` that is should stop. But here you are using it to determinate if `memif_event_thread` is stopped? why? The period when both are running is correlated but does not have to be the same. 

In my opinion the issue is that mtl_st22p_tx_session_stop() stops `tx_ctx->memif_event_thread` which in my opinion it should not do. This causes the thread to be cancelled twice, which is the bug that we want to fix.

A little off topic, but I would separate all logic connected with memif from mtl.c file and move it to shm_memif_*.c. All variables connected to memif should be in separate structure as well. the structure then can be inside ctx, but we need some separation.

Sakoram: This solution works because `mtl_st22p_tx_session_stop()` not only stops `tx_ctx->memif_event_thread` thread, but it also sets `tx_ctx->stop` to `true`, but it is not good fix.

Sakoram: the variable name `stop` is also a little bit undescriptive. It should be named e.g. stop_frame_thread, or possibly it could be in small structure connected to this thread with thread descriptor, mutexes, conditional variables etc. ( this don't need to be done in this PR, just commenting )

Mionsz: I am aware of this, and I assume that was the purpose of `stop` variable. It is not used elsewhere. I would agree on doing the more complex logic fixes in separate PR and leave this one as a bugfix only. I will add todo comments to the code.

Sakoram: please take a look at:  https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/main/media-proxy/src/rdma_session.c#L462

No complex logic has to be added, the *_session_stop() are just not needed for TX, as there is no thread to be stoped. We can do nothing in them, or remove them at all.

Mionsz: Done. The refactor for remaining methods must be included in separate PR.

Sakoram: I'm ok with changes for st22. In my opinion it is worth to fix other protocols, but if they work fine I guess it can wait (honestly I don't understand why they work, because they use the same code)

Sakoram (approved):
LGTM. with comment that we need to do the same for st20, st30 and st40

PR #202, opened 2024-09-23
Cherry pick bug fixes from main to release branch maint-24.09

Sakoram:
This PR includes the following changes:
- Fix how threads are canceled (Resources may not have been deallocated at all or deallocated while the thread was still running)
- Add timeout for memif_buffer_alloc(), but only for RDMA
- Remove unnecessary pacing from sender_app
- Increase ring size for RDMA
- Improve test-rdma.sh

- Print throughput in sample apps
- Create script to prepare env for RDMA
- Update README

### DIFF
--- a/media-proxy/src/mtl.c
+++ b/media-proxy/src/mtl.c
@@ -1060,7 +1060,7 @@ int rx_shm_deinit(rx_session_context_t* rx_ctx)
     }
 
     err = pthread_join(rx_ctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1096,7 +1096,7 @@ int tx_shm_deinit(tx_session_context_t* tx_ctx)
     }
 
     err = pthread_join(tx_ctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1132,7 +1132,7 @@ int rx_st22p_shm_deinit(rx_st22p_session_context_t* rx_ctx)
     }
 
     err = pthread_join(rx_ctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1168,7 +1168,7 @@ int tx_st22p_shm_deinit(tx_st22p_session_context_t* tx_ctx)
     }
 
     err = pthread_join(tx_ctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1204,7 +1204,7 @@ int rx_st30_shm_deinit(rx_st30_session_context_t* pctx)
     }
 
     err = pthread_join(pctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1240,7 +1240,7 @@ int tx_st30_shm_deinit(tx_st30_session_context_t* pctx)
     }
 
     err = pthread_join(pctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1281,7 +1281,7 @@ int rx_st40_shm_deinit(rx_st40_session_context_t* pctx)
     }
 
     err = pthread_join(pctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -1317,7 +1317,7 @@ int tx_st40_shm_deinit(tx_st40_session_context_t* pctx)
     }
 
     err = pthread_join(pctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
@@ -2946,7 +2946,7 @@ int rx_udp_h264_shm_deinit(rx_udp_h264_session_context_t* rx_ctx)
     }
 
     err = pthread_join(rx_ctx->memif_event_thread, NULL);
-    if (err) {
+    if (err && err != ESRCH) {
         ERROR("%s: Error joining thread: %s", __func__, strerror(err));
     }
 
--- a/media-proxy/src/mtl.c
+++ b/media-proxy/src/mtl.c
@@ -295,7 +295,7 @@ static int rx_st20p_query_ext_frame(void* priv, struct st_ext_frame* ext_frame,
         return -1;
     }
 
-    err = memif_buffer_alloc_timeout(rx_ctx->memif_conn, qid, rx_bufs, buf_num, &rx_buf_num, buf_size, 10);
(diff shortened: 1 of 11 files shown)


ko80 (approved):
Looks good to me

Mionsz (commented):
Overall seems ok - but what about segfault?

media-proxy/src/mtl.c
```diff
@@ -1125,12 +1155,22 @@ int rx_st22p_shm_deinit(rx_st22p_session_context_t* rx_ctx)
...
-    // pthread_cancel(tx_ctx->memif_event_thread);
+    err = pthread_cancel(tx_ctx->memif_event_thread);
+    if (err) {
+        ERROR("%s: Error canceling thread: %s", __func__, strerror(err));
+    }
+
+    err = pthread_join(tx_ctx->memif_event_thread, NULL);
+    if (err && err != ESRCH) {
+        ERROR("%s: Error joining thread: %s", __func__, strerror(err));
+    }
```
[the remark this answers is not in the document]
soopel: We will take care of segfault in the next cherry pick.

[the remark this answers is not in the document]
Sakoram: Your fix to that issue also is needed here. You are right that this change can introduce this problem to release branch. I guess better to wait for the fix.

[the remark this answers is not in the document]
Sakoram: I've added your fix

ko80 (approved):

media-proxy/src/shm_memif_common.c
```diff
@@ -196,3 +196,24 @@ int tx_on_disconnect(memif_conn_handle_t conn, void* priv_data)
...
+            waited_half_ms += 1;
+            continue;
+        }
+        if (err)
+            return err;
+        else
+            return 0;
+    }
+    return MEMIF_ERR_NOBUF_RING;
+}
```
Mionsz, line 219:
```suggestion
int memif_buffer_alloc_timeout(memif_conn_handle_t conn, uint16_t qid,
                               memif_buffer_t * bufs, uint16_t count, uint16_t * count_out,
                               uint32_t size, uint32_t timeout_ms) {
    timeout_ms = timeout_ms*2;
    int err;

    do {
        err = memif_buffer_alloc(conn, qid, bufs, count, count_out, size);
        if (err != MEMIF_ERR_NOBUF_RING) {
            return err;
         }
         usleep(500);
    } while (--timeout_ms)
    return MEMIF_ERR_NOBUF_RING;
}
```

Mionsz: Ony asking if this could be like above.

Mionsz (approved):
For me all is good. We need to test it for sure.

media-proxy/src/shm_memif_common.c
ko80: You're proposing to invert the logic of exiting the loop. Do you think this is easier to understand the code?

Mionsz: Ok you've got the point, this is less readable.

Sakoram:
Added new commits, which add:
- Print throughput in sample apps
- Create script to prepare env for RDMA
- Update README

ko80:
Adding a timeout in memif_buffer_alloc() can potentially break the flow of sending/receiving frames in sample apps and FFmpeg MCM plugins. This change need through review.

tests/single-node-sample-apps/test-rdma.sh
```diff
@@ -136,14 +144,14 @@ function run_test_rdma() {
...
     run_in_background "$bin_dir/$tx_media_proxy_cmd" "$tx_media_proxy_out"
     tx_media_proxy_pid=$!
 
     sleep 1
 
     info "Starting Rx side media_proxy"
-    local rx_media_proxy_cmd="media_proxy -i 192.168.96.2 -t 8003"
+    local rx_media_proxy_cmd="media_proxy -t 8003"
     run_in_background "$bin_dir/$rx_media_proxy_cmd" "$rx_media_proxy_out"
     rx_media_proxy_pid=$!
```
Mionsz, line 156: Hey I am not sure what you wanted to achieve but this is totally wrong. This way media proxy will default to predefined parameters and the MTL will be initialized for them as well. Currently the parameters that are mandatory are `-d` and `-i` and probably this should be reflected in not using defaults but exiting with error:

image

scripts/setup_rdma_env.sh
```diff
@@ -0,0 +1,162 @@
...
+        echo "FAIL: No RDMA devices detected"
+        exit 1
+    fi
+    fail=0
+    for device in "${rdma_devices[@]}"; do
+        mtu=$(ip addr | grep "$device:" | cut -d " " -f 5)
+        if [[ "$mtu" == "9000" ]]; then
+            echo "* $device OK"
+        else
+            echo "* $device FAIL: MTU = $mtu (9000 is recommended)"
```
Mionsz, line 106: If it is recommended then it should not be a fail but a warning.

scripts/setup_rdma_env.sh
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+WORKING_DIR="$(readlink -f "$(dirname -- "${BASH_SOURCE[0]}")")/rdma"
```
Mionsz, line 5:
```suggestion
SCRIPT_DIR="$(readlink -f "$(dirname -- "${BASH_SOURCE[0]}")")"
REPOSITORY_DIR="$(readlink -f "${SCRIPT_DIR}/..")"
WORKING_DIR="${REPOSITORY_DIR}/build/libfabric"
```

scripts/setup_rdma_env.sh
```diff
@@ -0,0 +1,162 @@
...
+    make -j$(nproc)
+    sudo make install
+    popd
+    popd
+}
+
+install_libfabric() {
+    echo "Installing libfabric from the source..."
+    pushd "$WORKING_DIR"
+    git clone --depth 1 --branch v1.22.0 https://github.com/ofiwg/libfabric libfabric
```
Mionsz, line 72: This is done wrongly - this will download the libfabric to your working directory (place from where you will call this script). It is also better to use the `wget` or `curl` and download it as `tar.gz` or just have it as a repository module.

`git clone --depth 1 --branch v1.22.0 https://github.com/ofiwg/libfabric ${WORKING_DIR}/libfabric`

or
```bash
. ${REPOSITORY_DIR}/common.sh

# GITHUB_CREDENTIALS="username:password"
# URL construction: https://${GITHUB_CREDENTIALS}@github.com/${name}/archive/${version}.tar.gz
# $1 - name
# $2 - version
# $3 - dest_dir
function git_download_strip_unpack()
{
    # Version can be commit sha or tag, examples:
    # version=d2515b90cc0ef651f6d0a6661d5a644490bfc3f3
    # version=refs/tags/v${JPEG_XS_VER}
    name="${1}"
    version="${2}"
    dest_dir="${3}"
    filename="$(get_filename ${version})"
    [ -n "${GITHUB_CREDENTIALS}" ] && creds="${GITHUB_CREDENTIALS}@" || creds=""

    mkdir -p "${dest_dir}"
    curl -Lf https://${creds}github.com/${name}/archive/${version}.tar.gz -o "${dest_dir}/${filename}.tar.gz"
    tar -zx --strip-components=1 -C "${dest_dir}" -f "${dest_dir}/${filename}.tar.gz"
    rm -f "${dest_dir}/${filename}.tar.gz"
}

LIBFABRIC_VER="v1.22.0"
git_download_strip_unpack "ofiwg/libfabric" "${LIBFABRIC_VER}" "${LIBFABRIC_DIR}"
```

tests/single-node-sample-apps/test-rdma.sh
Sakoram: this is only for RDMA. This test don't create any MTL session. The parameters are not used, so I did not pass any

scripts/setup_rdma_env.sh
Sakoram: agree, We should fix it on the main branch, but I think we don't have to fix it now.

PR #204, opened 2024-09-25
Print throughput in sample apps

Sakoram:


ko80 (commented):
Looks good

ko80 (approved):

Sakoram:
I checked the performance of st20 and memif with this change. No regression spotted.

PR #205, opened 2024-09-25
Fix last packet contents in sender_app

ko80:
Zero the whole buffer provided by MCM before reading data from a file. Fix the arguments order in fread().


MateuszGrabuszynski:
Works as expected. Top is padded output, bottom is input.
image

MateuszGrabuszynski (approved):
LGTM. As commented above.

PR #213, opened 2024-09-30
Refactor MCM Data Plane SDK API + upgrade FFmpeg plugins and sample apps

ko80:
* Refactor Mesh Data Plane SDK API.
* Upgrade sample apps to the new SDK API.
* Upgrade FFmpeg video and audio plugins to the new SDK API.
* Remove dependency on mcm_dp.h.
* Resolve API review discussions.
* Calculate RDMA transfer size before creating connection.

### DIFF
--- a/docs/sdk/SDK_API_DEFINITION.md
+++ b/docs/sdk/SDK_API_DEFINITION.md
@@ -1,0 +1,205 @@
+# Mesh Data Plane SDK Definition
+
+## mesh_create_client()
+```c
+int mesh_create_client(MeshClient **mc,
+                       MeshClientConfig *cfg)
+```
+Creates a new mesh client from the given configuration structure.
+
+#### Parameters
+* `[OUT]` `mc` – Address of a pointer to a mesh client structure.
+* `[IN]` `cfg` – Pointer to a mesh client configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_delete_client()
+```c
+int mesh_delete_client(MeshClient **mc)
+```
+Deletes the mesh client and its resources.
+
+#### Parameters
+* `[IN/OUT]` `mc` – Address of a pointer to a mesh client structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_create_connection()
+```c
+int mesh_create_connection(MeshClient *mc,
+                           MeshConnection **conn)
+```
+Creates a new media connection for the given mesh client.
+
+#### Parameters
+* `[IN]` `mc` – Pointer to a parent mesh client.
+* `[OUT]` `conn` – Address of a pointer to the connection structure.
+* `[IN]` `cfg` – Pointer to a connection configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_memif()
+```c
+int mesh_apply_connection_config_memif(MeshConnection *conn,
+                                       MeshConfig_Memif *cfg)
+```
+Applies the configuration to setup a single node direct connection via memif
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_st2110()
+```c
+int mesh_apply_connection_config_st2110(MeshConnection *conn,
+                                        MeshConfig_ST2110 *cfg)
+```
+Applies the configuration to setup an SMPTE ST2110-xx connection via Media Proxy
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_rdma()
+```c
+int mesh_apply_connection_config_rdma(MeshConnection *conn,
+                                      MeshConfig_RDMA *cfg)
+```
+Applies the configuration to setup an RDMA connection via Media Proxy
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_video()
+```c
+int mesh_apply_connection_config_video(MeshConnection *conn,
+                                       MeshConfig_Video *cfg)
+```
+Applies the configuration to setup the connection payload for Video frames
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_audio()
+```c
+int mesh_apply_connection_config_audio(MeshConnection *conn,
+                                       MeshConfig_Audio *cfg)
+```
+Applies the configuration to setup the connection payload for Audio packets
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_delete_connection()
+```c
+int mesh_delete_connection(MeshConnection **conn)
+```
+Deletes the connection and its resources.
+
+#### Parameters
+* `[IN/OUT]` `conn` – Address of a pointer to the connection structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_get_buffer()
+```c
+int mesh_get_buffer(MeshConnection *conn,
+                    MeshBuffer **buf)
+```
+Gets a buffer from the media connection.
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[OUT]` `buf` – Address of a pointer to a mesh buffer structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_get_buffer_timeout()
+```c
+int mesh_get_buffer_timeout(MeshConnection *conn,
+                            MeshBuffer **buf,
+                            int timeout_ms)
+```
+Gets a buffer from the media connection with a timeout.
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[OUT]` `buf` – Address of a pointer to a mesh buffer structure.
+* `[IN]` `timeout_ms` – Timeout interval in milliseconds.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_put_buffer()
+```c
+int mesh_put_buffer(MeshBuffer **buf)
+```
+Puts the buffer to the media connection.
+
+#### Parameters
+* `[IN/OUT]` `buf` – Address of a pointer to a mesh buffer structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_put_buffer_timeout()
+```c
+int mesh_put_buffer_timeout(MeshBuffer **buf,
+                            int timeout_ms)
+```
+Puts the buffer to the media connection with a timeout.
+
+#### Parameters
+* `[IN/OUT]` `buf` – Address of a pointer to a mesh buffer structure.
+* `[IN]` `timeout_ms` – Timeout interval in milliseconds.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_err2str()
+```c
+const char *mesh_err2str(int err)
+```
+Gets a text description of the error code.
+
+#### Parameters
+* `[IN]` `err` – Error code returned from any Mesh DP API call.
+
+#### Returns
+NULL-terminated string describing the error code.
--- a/docs/sdk/SDK_API_DEFINITION.md
+++ b/docs/sdk/SDK_API_DEFINITION.md
@@ -12,7 +12,7 @@ Creates a new mesh client from the given configuration structure.
 * `[IN]` `cfg` – Pointer to a mesh client configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_delete_client()
@@ -25,7 +25,7 @@ Deletes the mesh client and its resources.
 * `[IN/OUT]` `mc` – Address of a pointer to a mesh client structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_create_connection()
@@ -38,10 +38,9 @@ Creates a new media connection for the given mesh client.
 #### Parameters
 * `[IN]` `mc` – Pointer to a parent mesh client.
 * `[OUT]` `conn` – Address of a pointer to the connection structure.
-* `[IN]` `cfg` – Pointer to a connection configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_apply_connection_config_memif()
@@ -49,14 +48,14 @@ Creates a new media connection for the given mesh client.
 int mesh_apply_connection_config_memif(MeshConnection *conn,
                                        MeshConfig_Memif *cfg)
 ```
-Applies the configuration to setup a single node direct connection via memif
+Applies the configuration to setup a single node direct connection via memif.
 
 #### Parameters
 * `[IN]` `conn` – Pointer to a connection structure.
 * `[IN]` `cfg` – Pointer to a configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_apply_connection_config_st2110()
@@ -64,14 +63,14 @@ Applies the configuration to setup a single node direct connection via memif
 int mesh_apply_connection_config_st2110(MeshConnection *conn,
                                         MeshConfig_ST2110 *cfg)
 ```
-Applies the configuration to setup an SMPTE ST2110-xx connection via Media Proxy
+Applies the configuration to setup an SMPTE ST2110-xx connection via Media Proxy.
 
 #### Parameters
 * `[IN]` `conn` – Pointer to a connection structure.
 * `[IN]` `cfg` – Pointer to a configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_apply_connection_config_rdma()
@@ -79,14 +78,14 @@ Applies the configuration to setup an SMPTE ST2110-xx connection via Media Proxy
 int mesh_apply_connection_config_rdma(MeshConnection *conn,
                                       MeshConfig_RDMA *cfg)
 ```
-Applies the configuration to setup an RDMA connection via Media Proxy
+Applies the configuration to setup an RDMA connection via Media Proxy.
 
 #### Parameters
 * `[IN]` `conn` – Pointer to a connection structure.
 * `[IN]` `cfg` – Pointer to a configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_apply_connection_config_video()
@@ -94,14 +93,14 @@ Applies the configuration to setup an RDMA connection via Media Proxy
 int mesh_apply_connection_config_video(MeshConnection *conn,
                                        MeshConfig_Video *cfg)
 ```
-Applies the configuration to setup the connection payload for Video frames
+Applies the configuration to setup the connection payload for Video frames.
 
 #### Parameters
 * `[IN]` `conn` – Pointer to a connection structure.
 * `[IN]` `cfg` – Pointer to a configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_apply_connection_config_audio()
@@ -109,14 +108,42 @@ Applies the configuration to setup the connection payload for Video frames
 int mesh_apply_connection_config_audio(MeshConnection *conn,
                                        MeshConfig_Audio *cfg)
 ```
-Applies the configuration to setup the connection payload for Audio packets
+Applies the configuration to setup the connection payload for Audio packets.
 
 #### Parameters
 * `[IN]` `conn` – Pointer to a connection structure.
 * `[IN]` `cfg` – Pointer to a configuration structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
+
+
+## mesh_establish_connection()
+```c
+int mesh_establish_connection(MeshConnection *conn,
+                              int kind)
+```
+Applies the configuration to setup the connection payload for Audio packets.
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `kind` – Connection kind: Sender or Receiver.
+
+#### Returns
+0 on success; an error code otherwise.
+
+
+## mesh_shutdown_connection()
+```c
+int mesh_shutdown_connection(MeshConnection *conn)
+```
+Closes the active mesh connection.
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+
+#### Returns
+0 on success; an error code otherwise.
 
 
 ## mesh_delete_connection()
@@ -129,7 +156,7 @@ Deletes the connection and its resources.
 * `[IN/OUT]` `conn` – Address of a pointer to the connection structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_get_buffer()
@@ -144,7 +171,7 @@ Gets a buffer from the media connection.
 * `[OUT]` `buf` – Address of a pointer to a mesh buffer structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_get_buffer_timeout()
@@ -161,7 +188,7 @@ Gets a buffer from the media connection with a timeout.
 * `[IN]` `timeout_ms` – Timeout interval in milliseconds.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_put_buffer()
@@ -174,7 +201,7 @@ Puts the buffer to the media connection.
 * `[IN/OUT]` `buf` – Address of a pointer to a mesh buffer structure.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_put_buffer_timeout()
@@ -189,7 +216,7 @@ Puts the buffer to the media connection with a timeout.
 * `[IN]` `timeout_ms` – Timeout interval in milliseconds.
 
 #### Returns
-0 if successful. Otherwise, returns an error.
+0 on success; an error code otherwise.
 
 
 ## mesh_err2str()
@@ -199,7 +226,7 @@ const char *mesh_err2str(int err)
 Gets a text description of the error code.
 
 #### Parameters
-* `[IN]` `err` – Error code returned from any Mesh DP API call.
+* `[IN]` `err` – Error code returned from any Mesh Data Plane API call.
 
 #### Returns
 NULL-terminated string describing the error code.
--- a/docs/sdk/SDK_API_EXAMPLES.md
+++ b/docs/sdk/SDK_API_EXAMPLES.md
@@ -1,0 +1,216 @@
+# Mesh Data Plane SDK – Examples
+
+## SMPTE ST2110-22 compressed video sender example
+
+```c
+int main(void)
+{
+    /* Default client configuration */
+    MeshClientConfig client_config = { 0 };
+    
+    /* ST2110-XX configuration */
+    MeshConfig_ST2110 conn_config = {
+        .remote_ip_addr = "192.168.96.2",
+        .remote_port = 9001,
+        .transport = MESH_CONN_TRANSPORT_ST2110_22,
+    };
+
+    /* Video configuration */
+    MeshConfig_Video payload_config = {
+        .width = 1920,
+        .height = 1080,
+        .fps = 60,
+        .pixel_format = MESH_VIDEO_PIXEL_FORMAT_YUV422P10LE,
+    };
+
+    MeshConnection *conn;
+    MeshClient *mc;
+    int err;
+    int i, n;
+
+    /* Create a mesh client */
+    err = mesh_create_client(&mc, &client_config);
+    if (err) {
+        printf("Failed to create mesh client: %s (%d)\n", mesh_err2str(err), err);
+        exit(1);
+    }
+
+    /* Create a mesh connection */
+    err = mesh_create_connection(mc, &conn);
+    if (err) {
+        printf("Failed to create connection: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_client;
+    }
+
+    /* Apply connection configuration */
+    err = mesh_apply_connection_config_st2110(conn, &conn_config);
+    if (err) {
+        printf("Failed to apply SMPTE ST2110 configuration: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* Apply video payload configuration */
+    err = mesh_apply_connection_config_video(conn, &video_config);
+    if (err) {
+        printf("Failed to apply video configuration: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* Establish a connection for sending data */
+    err = mesh_establish_connection(conn, MESH_CONN_KIND_SENDER);
+    if (err) {
+        printf("Failed to establish connection: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* 10 video frames to be sent */
+    n = 10;
+
+    /* Send data loop */
+    for (i = 0; i < n; i++) {
+        MeshBuffer *buf;
+
+        /* Ask the mesh to allocate a shared memory buffer for user data */
+        err = mesh_get_buffer(conn, &buf);
+        if (err) {
+            printf("Failed to get buffer: %s (%d)\n", mesh_err2str(err), err);
+            break;
+        }
+
+        /* Fill the buffer with user data */
+        put_user_video_frames(buf->data, buf->data_len);
+
+        /* Send the buffer */
+        err = mesh_put_buffer(&buf);
+        if (err) {
+            printf("Failed to put buffer: %s (%d)\n", mesh_err2str(err), err);
+            break;
+        }
+    }
+
+    /* Shutdown the connection */
+    err = mesh_shutdown_connection(&conn);
+    if (err)
+        printf("Failed to shutdown connection: %s (%d)\n", mesh_err2str(err), err);
+
+exit_delete_conn:
+    /* Delete the media connection */
+    mesh_delete_connection(&conn);
+
+exit_delete_client:
+    /* Delete the mesh client */
+    mesh_delete_client(&mc);
+
+    if (err)
+        exit(1);
+}
+```
+
+## SMPTE ST2110-22 compressed video receiver example
+```c
+int main(void)
+{
+    /* Default client configuration */
+    MeshClientConfig client_config = { 0 };
+    
+    /* ST2110-XX configuration */
+    MeshConfig_ST2110 conn_config = {
+        .remote_ip_addr = "192.168.96.1",
+        .local_port = 9001,
+        .transport = MESH_CONN_TRANSPORT_ST2110_22,
+    };
+
+    /* Video configuration */
+    MeshConfig_Video payload_config = {
+        .width = 1920,
+        .height = 1080,
+        .fps = 60,
+        .pixel_format = MESH_VIDEO_PIXEL_FORMAT_YUV422P10LE,
+    };
+
+    MeshConnection *conn;
+    MeshClient *mc;
+    int err;
+    int i, n;
+
+    /* Create a mesh client */
+    err = mesh_create_client(&mc, &client_config);
+    if (err) {
+        printf("Failed to create mesh client: %s (%d)\n", mesh_err2str(err), err);
+        exit(1);
+    }
+
+    /* Create a mesh connection */
+    err = mesh_create_connection(mc, &conn);
+    if (err) {
+        printf("Failed to create connection: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_client;
+    }
+
+    /* Apply connection configuration */
+    err = mesh_apply_connection_config_st2110(conn, &conn_config);
+    if (err) {
+        printf("Failed to apply SMPTE ST2110 configuration: %s (%d)\n",
+               mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* Apply video payload configuration */
+    err = mesh_apply_connection_config_video(conn, &video_config);
+    if (err) {
+        printf("Failed to apply video configuration: %s (%d)\n",
+               mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* Establish a connection for receiving data */
+    err = mesh_establish_connection(conn, MESH_CONN_KIND_RECEIVER);
+    if (err) {
+        printf("Failed to establish connection: %s (%d)\n",
+               mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* Receive data loop */
+    do {
+        MeshBuffer *buf;
+
+        /* Receive a buffer from the mesh */
+        err = mesh_get_buffer(conn, &buf);
+        if (err == MESH_ERR_CONNECTION_CLOSED) {
+            printf("Connection closed\n");
+            break;
+        }
+        if (err) {
+            printf("Failed to get buffer: %s (%d)\n", mesh_err2str(err), err);
+            break;
+        }
+
+        /* Process the received user data */
+        get_user_video_frames(buf->data, buf->data_len);
+
+        /* Release and put the buffer back to the mesh */
+        err = mesh_put_buffer(&buf);
+        if (err) {
+            printf("Failed to put buffer: %s (%d)\n", mesh_err2str(err), err);
+            break;
+        }
+    } while (1);
+
+    /* Shutdown the connection */
+    err = mesh_shutdown_connection(&conn);
+    if (err)
+        printf("Failed to shutdown connection: %s (%d)\n", mesh_err2str(err), err);
+
+exit_delete_conn:
+    /* Delete the media connection */
+    mesh_delete_connection(&conn);
+
+exit_delete_client:
+    /* Delete the mesh client */
+    mesh_delete_client(&mc);
+
+    if (err)
+        exit(1);
+}
+```
--- a/docs/sdk/SDK_API_EXAMPLES.md
+++ b/docs/sdk/SDK_API_EXAMPLES.md
@@ -24,6 +24,7 @@ int main(void)
     };
 
     MeshConnection *conn;
+    MeshBuffer *buf;
     MeshClient *mc;
     int err;
     int i, n;
@@ -68,8 +69,6 @@ int main(void)
 
     /* Send data loop */
     for (i = 0; i < n; i++) {
-        MeshBuffer *buf;
-
         /* Ask the mesh to allocate a shared memory buffer for user data */
         err = mesh_get_buffer(conn, &buf);
         if (err) {
--- a/docs/sdk/SDK_API_WORKFLOW.md
+++ b/docs/sdk/SDK_API_WORKFLOW.md
@@ -1,0 +1,170 @@
+# Mesh Data Plane SDK – API Workflow
+## General workflow
+1. Create a Mesh client:
+   * `mesh_create_client()`
+1. Create a Mesh connection:
+   * `mesh_create_connection()`
+1. Apply the connection configuration:
+   * `struct MeshConfig_Memif {...}` -> `mesh_apply_config_memif()`
+   * `struct MeshConfig_ST2110 {...}` -> `mesh_apply_config_st2110()`
+   * `struct MeshConfig_RDMA {...}` -> `mesh_apply_config_rdma()`
+1. Apply the payload configuration:
+   * `struct MeshConfig_Video` -> `mesh_apply_config_video()`
+   * `struct MeshConfig_Audio` -> `mesh_apply_config_audio()`
+1. Establish the Mesh connection:
+   * `mesh_establish_connection()`
+1. Get a buffer from the Mesh connection:
+   * `mesh_get_buffer()`
+   * `mesh_get_buffer_timeout()`
+1. Put the buffer to the Mesh connection:
+   * `mesh_put_buffer()`
+   * `mesh_put_buffer_timeout()`
+1. Shutdown the Mesh connection:
+   * `mesh_shutdown_connection()`
+1. Delete the Mesh connection:
+   * `mesh_delete_connection()`
+1. Delete the Mesh client:
+   * `mesh_delete_client()`
+
+## Configure connection – Memif
+1. Define a memif configuration:
+   * `struct MeshConfig_Memif`
+      * `.socket_path`
+      * `.interface_id`
+1. Apply the configuration:
+   * `mesh_apply_config_memif()`
+
+## Configure connection – SMPTE ST2110-XX
+1. Define an SMPTE ST2110-XX configuration:
+   * `struct MeshConfig_ST2110`
+      * `.remote_ip_addr`
+      * `.remote_port`
+      * `.local_ip_addr`
+      * `.local_port`
+      * `.transport`
+         * `MESH_CONN_TRANSPORT_ST2110_20` – SMPTE ST2110-20 Uncompressed Video
+         * `MESH_CONN_TRANSPORT_ST2110_22` – SMPTE ST2110-22 Constant Bit-Rate Compressed Video
+         * `MESH_CONN_TRANSPORT_ST2110_30` – SMPTE ST2110-30 Audio
+1. Apply the configuration:
+   * `mesh_apply_config_st2110()`
+
+## Configure connection – RDMA
+1. Define an RDMA configuration:
+   * `struct MeshConfig_RDMA`
+      * `.remote_ip_addr`
+      * `.remote_port`
+      * `.local_ip_addr`
+      * `.local_port`
+1. Apply the configuration:
+   * `mesh_apply_config_rdma()`
+
+## Configure payload – Video
+1. Define a video configuration:
+   * `struct MeshConfig_Video`
+      * `.width`
+      * `.height`
+      * `.fps`
+      * `.pixel_format` – `MESH_VIDEO_PIXEL_FORMAT_*`
+1. Apply the configuration:
+   * `mesh_apply_config_video()`
+
+## Configure payload – Audio
+1. Define an audio configuration:
+   * `struct MeshConfig_Audio`
+      * `.channels`
+      * `.sample_rate` – `MESH_AUDIO_SAMPLE_RATE_*`
+      * `.format` – `MESH_AUDIO_FORMAT_*`
+      * `.packet_time` – `MESH_AUDIO_PACKET_TIME_*`
+1. Apply the configuration:
+   * `mesh_apply_config_audio()`
+
+## Mesh DP SDK API internals – Connection configuration flowchart
+
+```mermaid
+flowchart LR
+   kind(.kind) --> sender[_Sender_]
+   kind --> receiver[_Receiver_]
+
+   conn_type(.conn_type) --> memif_t[_memif_]
+   conn_type --> st_t[_SMPTE ST2110-XX_]
+   conn_type --> rdma_t[_RDMA_]
+
+   conn(.conn) --> memif(.memif)
+      memif --> socket_path(".socket_path
+                             .interface_id")
+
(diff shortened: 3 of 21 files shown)


ko80:
Please note unit tests have passed
```
[==========] Running 29 tests from 5 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from APITests_MeshClient
[ RUN      ] APITests_MeshClient.Test_CreateDeleteClient
[       OK ] APITests_MeshClient.Test_CreateDeleteClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_CreateClient
[       OK ] APITests_MeshClient.TestNegative_CreateClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_DeleteClient
[       OK ] APITests_MeshClient.TestNegative_DeleteClient (0 ms)
[----------] 3 tests from APITests_MeshClient (0 ms total)

[----------] 18 tests from APITests2_MeshConnection
[ RUN      ] APITests2_MeshConnection.Test_CreateEstablishShutdownDeleteConnection
[       OK ] APITests2_MeshConnection.Test_CreateEstablishShutdownDeleteConnection (100 ms)
[ RUN      ] APITests2_MeshConnection.Test_ApplyConfig
[       OK ] APITests2_MeshConnection.Test_ApplyConfig (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ApplyConfig
[       OK ] APITests2_MeshConnection.TestNegative_ApplyConfig (0 ms)
[ RUN      ] APITests2_MeshConnection.Test_ParseConnectionConfig
[       OK ] APITests2_MeshConnection.Test_ParseConnectionConfig (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ParseConnectionConfigInval
[       OK ] APITests2_MeshConnection.TestNegative_ParseConnectionConfigInval (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ParseConnectionConfigIncompat
[       OK ] APITests2_MeshConnection.TestNegative_ParseConnectionConfigIncompat (0 ms)
[ RUN      ] APITests2_MeshConnection.Test_ParseVideoPayloadConfig
[       OK ] APITests2_MeshConnection.Test_ParseVideoPayloadConfig (0 ms)
[ RUN      ] APITests2_MeshConnection.Test_ParseAudioPayloadConfig
[       OK ] APITests2_MeshConnection.Test_ParseAudioPayloadConfig (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ParsePayloadConfigInval
[       OK ] APITests2_MeshConnection.TestNegative_ParsePayloadConfigInval (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ParsePayloadConfigIncompat
[       OK ] APITests2_MeshConnection.TestNegative_ParsePayloadConfigIncompat (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_EstablishConnectionConfigInval
[       OK ] APITests2_MeshConnection.TestNegative_EstablishConnectionConfigInval (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_CreateConnectionConfigIncompat
[       OK ] APITests2_MeshConnection.TestNegative_CreateConnectionConfigIncompat (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn
[       OK ] APITests2_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_CreateConnection_NulledConn
[       OK ] APITests2_MeshConnection.TestNegative_CreateConnection_NulledConn (0 ms)
[ RUN      ] APITests2_MeshConnection.Test_CreateConnection_MaxConnNumber
[       OK ] APITests2_MeshConnection.Test_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_CreateConnection_MaxConnNumber
[       OK ] APITests2_MeshConnection.TestNegative_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_ShutdownConnection_NulledConn
[       OK ] APITests2_MeshConnection.TestNegative_ShutdownConnection_NulledConn (0 ms)
[ RUN      ] APITests2_MeshConnection.TestNegative_DeleteConnection_NulledConn
[       OK ] APITests2_MeshConnection.TestNegative_DeleteConnection_NulledConn (0 ms)
[----------] 18 tests from APITests2_MeshConnection (100 ms total)

[----------] 6 tests from APITests2_MeshBuffer
[ RUN      ] APITests2_MeshBuffer.Test_GetPutBuffer
[       OK ] APITests2_MeshBuffer.Test_GetPutBuffer (50 ms)
[ RUN      ] APITests2_MeshBuffer.Test_GetBufferConnClosed
[       OK ] APITests2_MeshBuffer.Test_GetBufferConnClosed (50 ms)
[ RUN      ] APITests2_MeshBuffer.Test_GetBufferDefaultTimeout
[       OK ] APITests2_MeshBuffer.Test_GetBufferDefaultTimeout (100 ms)
[ RUN      ] APITests2_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf
[       OK ] APITests2_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf (0 ms)
[ RUN      ] APITests2_MeshBuffer.TestNegative_GetBuffer_NulledBuf
[       OK ] APITests2_MeshBuffer.TestNegative_GetBuffer_NulledBuf (50 ms)
[ RUN      ] APITests2_MeshBuffer.TestNegative_PutBuffer_NulledBuf
[       OK ] APITests2_MeshBuffer.TestNegative_PutBuffer_NulledBuf (0 ms)
[----------] 6 tests from APITests2_MeshBuffer (250 ms total)

[----------] 1 test from APITests_MeshBuffer
[ RUN      ] APITests_MeshBuffer.Test_ImportantConstants
[       OK ] APITests_MeshBuffer.Test_ImportantConstants (0 ms)
[----------] 1 test from APITests_MeshBuffer (0 ms total)

[----------] 1 test from ProxyContextTests
[ RUN      ] ProxyContextTests.ProxyContextConstructor
[       OK ] ProxyContextTests.ProxyContextConstructor (0 ms)
[----------] 1 test from ProxyContextTests (0 ms total)

[----------] Global test environment tear-down
[==========] 29 tests from 5 test suites ran. (350 ms total)
[  PASSED  ] 29 tests.
```

ko80:
NOTE: unit tests are all passing
```
Running main() from ./googletest/src/gtest_main.cc
[==========] Running 29 tests from 4 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from APITests_MeshClient
[ RUN      ] APITests_MeshClient.Test_CreateDeleteClient
[       OK ] APITests_MeshClient.Test_CreateDeleteClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_CreateClient
[       OK ] APITests_MeshClient.TestNegative_CreateClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_DeleteClient
[       OK ] APITests_MeshClient.TestNegative_DeleteClient (0 ms)
[----------] 3 tests from APITests_MeshClient (0 ms total)

[----------] 18 tests from APITests_MeshConnection
[ RUN      ] APITests_MeshConnection.Test_CreateEstablishShutdownDeleteConnection
[       OK ] APITests_MeshConnection.Test_CreateEstablishShutdownDeleteConnection (100 ms)
[ RUN      ] APITests_MeshConnection.Test_ApplyConfig
[       OK ] APITests_MeshConnection.Test_ApplyConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ApplyConfig
[       OK ] APITests_MeshConnection.TestNegative_ApplyConfig (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseConnectionConfig
[       OK ] APITests_MeshConnection.Test_ParseConnectionConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParseConnectionConfigInval
[       OK ] APITests_MeshConnection.TestNegative_ParseConnectionConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParseConnectionConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_ParseConnectionConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseVideoPayloadConfig
[       OK ] APITests_MeshConnection.Test_ParseVideoPayloadConfig (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseAudioPayloadConfig
[       OK ] APITests_MeshConnection.Test_ParseAudioPayloadConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParsePayloadConfigInval
[       OK ] APITests_MeshConnection.TestNegative_ParsePayloadConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParsePayloadConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_ParsePayloadConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_EstablishConnectionConfigInval
[       OK ] APITests_MeshConnection.TestNegative_EstablishConnectionConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnectionConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_CreateConnectionConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_NulledConn (0 ms)
[ RUN      ] APITests_MeshConnection.Test_CreateConnection_MaxConnNumber
[       OK ] APITests_MeshConnection.Test_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_MaxConnNumber
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ShutdownConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_ShutdownConnection_NulledConn (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_DeleteConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_DeleteConnection_NulledConn (0 ms)
[----------] 18 tests from APITests_MeshConnection (100 ms total)

[----------] 7 tests from APITests_MeshBuffer
[ RUN      ] APITests_MeshBuffer.Test_GetPutBuffer
[       OK ] APITests_MeshBuffer.Test_GetPutBuffer (50 ms)
[ RUN      ] APITests_MeshBuffer.Test_GetBufferConnClosed
[       OK ] APITests_MeshBuffer.Test_GetBufferConnClosed (50 ms)
[ RUN      ] APITests_MeshBuffer.Test_GetBufferDefaultTimeout
[       OK ] APITests_MeshBuffer.Test_GetBufferDefaultTimeout (100 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf
[       OK ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf (0 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledBuf
[       OK ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledBuf (50 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_PutBuffer_NulledBuf
[       OK ] APITests_MeshBuffer.TestNegative_PutBuffer_NulledBuf (0 ms)
[ RUN      ] APITests_MeshBuffer.Test_ImportantConstants
[       OK ] APITests_MeshBuffer.Test_ImportantConstants (0 ms)
[----------] 7 tests from APITests_MeshBuffer (250 ms total)

[----------] 1 test from ProxyContextTests
[ RUN      ] ProxyContextTests.ProxyContextConstructor
[       OK ] ProxyContextTests.ProxyContextConstructor (0 ms)
[----------] 1 test from ProxyContextTests (0 ms total)

[----------] Global test environment tear-down
[==========] 29 tests from 4 test suites ran. (351 ms total)
[  PASSED  ] 29 tests.
```

tests/unit/mesh_dp_api_tests.cc
```diff
@@ -0,0 +1,1490 @@
...
+int mock_enqueue_buf(mcm_conn_context *pctx, mcm_buffer *buf)
+{
+    free(buf);
+    return 0;
+}
+
+/**
+ * APITests setup
+ */
+void APITests_Setup()
```
Sakoram, line 98: Please take a look at https://github.com/meekrosoft/fff and consider using it. This would make it much easier to check things like whether the mock function has been called.

sdk/include/mesh_dp.h
```diff
@@ -0,0 +1,352 @@
...
+     */
+    const size_t buf_size;
+
+} MeshConnection;
+
+/**
+ * Define configuration string field sizes
+ */
+#define MESH_SOCKET_PATH_SIZE   108
+#define MESH_IP_ADDRESS_SIZE    46
```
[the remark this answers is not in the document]
ko80: Agree

sdk/include/mesh_dp_internal.h
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Memory copy function definition
+ */
+#define mesh_memcpy(dst, src, size) memcpy(dst, src, size)
+
+/**
+ * String size-bounded copy function definition
+ */
+#define mesh_strlcpy(dst, src, size) strlcpy(dst, src, size)
```
tszumski, line 108: I would suggest not to introduce wrappers like mesh_calloc() etc. This can make confusion "why author of that code cannot use basic method"

sdk/include/mesh_dp.h
```diff
@@ -0,0 +1,352 @@
...
+     * Any value of the MESH_AUDIO_PACKET_TIME_* constants.
+     */
+    int packet_time;
+/** Constants for 48kHz and 96kHz sample rates */
+#define MESH_AUDIO_PACKET_TIME_1MS      0 ///< Audio packet time 1ms
+#define MESH_AUDIO_PACKET_TIME_125US    1 ///< Audio packet time 125us
+#define MESH_AUDIO_PACKET_TIME_250US    2 ///< Audio packet time 250us
+#define MESH_AUDIO_PACKET_TIME_333US    3 ///< Audio packet time 333us
+#define MESH_AUDIO_PACKET_TIME_4MS      4 ///< Audio packet time 4ms
+#define MESH_AUDIO_PACKET_TIME_80US     5 ///< Audio packet time 80us
```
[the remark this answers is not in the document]
ko80: I would like to keep MESH_AUDIO_PACKET_TIME_1MS be equal to 0

sdk/include/mesh_dp_internal.h
ko80: The idea behind these definitions was to create an isolation layer for future optimizations. I remember there were suggestions to use optimized memory operations related functions from the MTL repo.

sdk/include/mesh_dp.h
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * Video frames per second.
+     */
+    double fps;
+    /**
+     * Video frame pixel format.
+     * Any value of the MESH_VIDEO_PIXEL_FORMAT_* constants.
+     */
+    int pixel_format;
+#define MESH_VIDEO_PIXEL_FORMAT_NV12        0 ///< planar YUV 4:2:0, 12bpp
```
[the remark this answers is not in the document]
ko80: The idea was to avoid introducing unnecessary type definitions for the user. Just use `int`. This concept is used in the FFmpeg code base. And the other reason to adopt that approach was to make the options visible along with the structure field definition.

These macros are only used as options of the structure field definition above them. Adding an `enum` will move their definitions far from the structure field, which will make it harder to understand by the user since they will need to scroll the code back and forth.

I agree, it's not an ideal approach, but I liked so much how similar structure definitions with macros bring clarity of the relationship between the field and the options in FFmpeg codebase. So I decided to adopt that.

Reference: https://github.com/FFmpeg/FFmpeg/blob/b9145fcab27da9e14c6929036d771afd1907a771/libavformat/avformat.h#L1302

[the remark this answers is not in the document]
Sakoram: And what about the case when the same enums/defines are used in multiple places? Should we define them twice and try to not include header file in which they were defined previously? or define just ones and lose benefit of them being defined right above the field definition?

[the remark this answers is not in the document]
ko80: I'd go with `enum` if the compiler doesn't allow you to do the below. **But it does**. GCC treats `enum` as an `int` and prints no warnings/errors even with `-Wall -Wenum_compare`.
```
enum MeshVideoPixelFormat {
    MESH_VIDEO_PIXEL_FORMAT_NV12        = 0, ///< planar YUV 4:2:0, 12bpp
    MESH_VIDEO_PIXEL_FORMAT_YUV422P     = 1, ///< planar YUV 4:2:2, 16bpp
    MESH_VIDEO_PIXEL_FORMAT_YUV422P10LE = 2, ///< planar YUV 4:2:2, 20bpp
    MESH_VIDEO_PIXEL_FORMAT_YUV444P10LE = 3, ///< planar YUV 4:4:4, 30bpp
    MESH_VIDEO_PIXEL_FORMAT_RGB8        = 4, ///< packed RGB 3:3:2,  8bpp
};

enum MeshVideoPixelFormat pixel_format;

...

pixel_format = 123;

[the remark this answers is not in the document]
Comment (ko80) [reply sdk/include/mesh_dp.h:168]: These macro constants have a four-level namespace prefix to make them look specifically related to the MCM library. Do you think the user will be defining something like `MESH_VIDEO_PIXEL_FORMAT_YUV422P10LE` in their code?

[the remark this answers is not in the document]
Comment (Sakoram) [reply sdk/include/mesh_dp.h:168]: ahh right, I forgot that only C++ prohibits this. Ok, type safety was my strongest argument. I would have chosen enum anyway, but it's mostly an opinion now. I don't like how struct filled with definitions looks, how it has broken indentation, but if you think that's better, We can do it this way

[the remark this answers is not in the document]
Comment (ko80) [reply sdk/include/mesh_dp.h:168]: Thank you. I'll collect all the feedback from reviewers to decide if we go with macros or refactor to `enum`.

File: ffmpeg-plugin/mcm_video_tx.c [hunk 6]
```diff
@@ -29,85 +28,102 @@ typedef struct McmVideoMuxerContext {
...
     }
 
     av_log(avctx, AV_LOG_INFO,
            "w:%d h:%d pixfmt:%s fps:%d\n",
            s->width, s->height, av_get_pix_fmt_name(s->pixel_format),
            (int)av_q2d(s->frame_rate));
     return 0;
+
+exit_delete_conn:
+    mesh_delete_connection(&s->conn);
```
moleksy, line 100: Shouldn't we check the output of mesh_delete_connection here?

ko80: This function call is already in the path of exiting due to an error happened before. So, the extra error checking here is redundant and could potentially be misleading as to the actual error occurred in the previous place in the code. I suggest leaving it here as it is.

tests/unit/mesh_dp_api_tests.cc
```diff
@@ -0,0 +1,1490 @@
...
+
+    APITests_Setup();
+
+    err = mesh_create_client(&mc, NULL);
+    ASSERT_EQ(err, 0) << mesh_err2str(err);
+    ASSERT_NE(mc, (MeshClient *)NULL);
+
+    err = mesh_create_connection(mc, &conn);
+    EXPECT_EQ(err, 0) << mesh_err2str(err);
+    EXPECT_NE(conn, (MeshConnection *)NULL);
```
moleksy, line 151: Possibly redundant, since if err == 0, conn cannot be null.

sdk/src/mesh_dp.c
```diff
@@ -0,0 +1,779 @@
...
+            break;
+        case MESH_VIDEO_PIXEL_FORMAT_RGB8:
+            param->pix_fmt = PIX_FMT_RGB8;
+            break;
+        default:
+            return -MESH_ERR_CONN_CONFIG_INVAL;
+        }
+
+        param->payload_args.video_args.pix_fmt = param->pix_fmt;
+        param->payload_args.video_args.width   = conn->cfg.payload.video.width;
```
Sakoram, line 322: it won't work for RDMA as it currently needs payload_args.rdma_args filled. As agreed, we can change rdma connection to take video_args, but it has to be done..

sdk/src/mesh_dp.c
```diff
@@ -0,0 +1,779 @@
...
+
+    mc_ctx = (MeshClientContext *)(*mc);
+
+    if (!mc_ctx)
+        return -MESH_ERR_BAD_CLIENT_PTR;
+
+    pthread_mutex_destroy(&mc_ctx->mx);
+
+    /**
+     * TODO: Shutdown and deallocate connections here
```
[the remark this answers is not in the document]
ko80: It is the user's responsibility to call `mesh_connection_delete()` to deallocate the connection. The user had to do that in the previous PR, and I'm not changing this flow.  This `TODO` has been left here for the future to remind us to implement the automatic connections shutdown and deallocation in `mesh_delete_client()` even when the user forgets to do that. So far it can't be implemented here because it requires additional mechanisms to gracefully shutdown the connection and synchronize everything. This is a future task.

sdk/src/mesh_dp.c
ko80: Good catch. Could you add a suggestion in this comment on what fields you'd like to populate?

docs/sdk/SDK_SMPTE_ST2110_RX_FLOW.md
```diff
@@ -0,0 +1,104 @@
+# Mesh Data Plane SDK – Example / SMPTE ST2110-22 Rx Flow
+
+The diagram in this document shows interaction between a user app and Mesh Data Plane in the video receive mode over SMPTE ST2110-22.
+
+### Rx flow stages
+1. Create client.
+1. Create connection.
+1. Configure connection.
+1. Establish Tx connection.
```
MateuszGrabuszynski, line 9: Should that not be Rx, instead of Tx?

MateuszGrabuszynski (commented):
I was able to review ONLY the `*.md` files before my leave. As there are a few comments, I put them here as `Comment` (not as `Request changes`) to avoid blocking the merge further down the line.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+```c
+int mesh_create_connection(MeshClient *mc,
+                           MeshConnection **conn)
+```
+Creates a new media connection for the given mesh client.
+
+#### Parameters
+* `[IN]` `mc` – Pointer to a parent mesh client.
+* `[OUT]` `conn` – Address of a pointer to the connection structure.
+* `[IN]` `cfg` – Pointer to a connection configuration structure.
```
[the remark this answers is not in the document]
ko80: Oh, my fault. Thanks for finding that :)

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+```
+Applies the configuration to setup the connection payload for Audio packets
+
+#### Parameters
+* `[IN]` `conn` – Pointer to a connection structure.
+* `[IN]` `cfg` – Pointer to a configuration structure.
+
+#### Returns
+0 if successful. Otherwise, returns an error.
+
```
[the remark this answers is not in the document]
ko80: That's for the future. For now the API is going to expose only tested features. We'll enable ST40 as soon as we add full support and test it.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+
+## mesh_put_buffer()
+```c
+int mesh_put_buffer(MeshBuffer **buf)
+```
+Puts the buffer to the media connection.
+
+#### Parameters
+* `[IN/OUT]` `buf` – Address of a pointer to a mesh buffer structure.
```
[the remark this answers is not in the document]
ko80: Then the buffer is put back to MCM, the ownership of the buffer is passed from the user to MCM. And MCM will nullify this pointer in `mesh_put_buffer()` to assure the user doesn't know it. So, this is an IN and OUT argument.

docs/sdk/SDK_API_EXAMPLES.md
```diff
@@ -0,0 +1,216 @@
...
+        printf("Failed to establish connection: %s (%d)\n", mesh_err2str(err), err);
+        goto exit_delete_conn;
+    }
+
+    /* 10 video frames to be sent */
+    n = 10;
+
+    /* Send data loop */
+    for (i = 0; i < n; i++) {
+        MeshBuffer *buf;
```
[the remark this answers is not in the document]
ko80: Well, from the resulting assembly code perspective, you won't see any difference if you move the pointer declaration upper. It will even sit in the same memory address on the stack.

There is no `creation` of the pointer in this block. It's just a declaration of a local variable on the stack which disappears when the execution flow exits the block. No memory leakage is possible here.

The only concern I could imagine here is the software developer's habit to declare local variables. From the C standard perspective, it is correct.

But yes, we can move it upper if you'd like.

docs/sdk/SDK_SMPTE_ST2110_RX_FLOW.md
ko80: Yep, my fault. Will update it.

sdk/src/mesh_dp.c
Sakoram: Here there is nothing to add, but we have to interpret those parameters differently in [ProxyContext::RxStart_rdma](https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/main/media-proxy/src/proxy_context.cc#L802) and  [ProxyContext::TxStart_rdma](https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/main/media-proxy/src/proxy_context.cc#L998)

sdk/include/mesh_dp_internal.h
Sakoram: I agree with Tomek. Somebody will waste its time while reading the code to check what is underneath the wrapper, but there is nothing and plain calloc() could be used.

If there will be some optimalization underneath we can add the wrapper then

sdk/src/mesh_dp.c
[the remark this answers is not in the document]
Sakoram: So currently we will have a memory leak if the user don't use mesh_delete_client() for every connection?

Is it really that complicated to implement this feature? It seems that traversing though the list and calling something like mesh_delete_client on every node will be sufficient.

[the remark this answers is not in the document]
ko80: No, the user must call `mesh_delete_client()` only once after deleting all connections with `mesh_delete_connection()`.

I believe we don't want the user to call `mesh_delete_client()` while the connection is still active in a parallel thread receiving or sending data. We don't want that to happen because there are no mechanisms of shutdown synchronization between the client and the connection. Before deleting a connection, the client must raise a signal to notify other threads that the connection and its resources are about to destroy. Then, every thread should react to this notification by closing the connection and decreasing the refcount. After that, when the connection refcount is decreased down to zero, the client can finally delete the connection.

This sort of a mechanism could be developed in the future as a new feature. For now, it's the user's responsibility to delete all connections before deleting the client.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
+# Mesh Data Plane SDK Definition
+
+## mesh_create_client()
+```c
+int mesh_create_client(MeshClient **mc,
+                       MeshClientConfig *cfg)
+```
+Creates a new mesh client from the given configuration structure.
+
+#### Parameters
```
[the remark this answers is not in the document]
ko80: Agree, it should sit in `mesh_dp.h` and I'm working on that ;-)

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+* `[OUT]` `mc` – Address of a pointer to a mesh client structure.
+* `[IN]` `cfg` – Pointer to a mesh client configuration structure.
```
skolelis, line 12: I think out/in "prefix" can be shown also in the function declaration itself in the code.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_memif()
+```c
+int mesh_apply_connection_config_memif(MeshConnection *conn,
+                                       MeshConfig_Memif *cfg)
+```
+Applies the configuration to setup a single node direct connection via memif
```
skolelis, line 52: "." at the end.

docs/sdk/SDK_API_DEFINITION.md
ko80: Agree. This is what I'm working on.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_rdma()
+```c
+int mesh_apply_connection_config_rdma(MeshConnection *conn,
+                                      MeshConfig_RDMA *cfg)
+```
+Applies the configuration to setup an RDMA connection via Media Proxy
```
skolelis, line 82: "." at the end of sentence.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_st2110()
+```c
+int mesh_apply_connection_config_st2110(MeshConnection *conn,
+                                        MeshConfig_ST2110 *cfg)
+```
+Applies the configuration to setup an SMPTE ST2110-xx connection via Media Proxy
```
skolelis, line 67: "." at the end of sentence.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_video()
+```c
+int mesh_apply_connection_config_video(MeshConnection *conn,
+                                       MeshConfig_Video *cfg)
+```
+Applies the configuration to setup the connection payload for Video frames
```
skolelis, line 97: "." at the end of sentence.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+#### Returns
+0 if successful. Otherwise, returns an error.
+
+
+## mesh_apply_connection_config_audio()
+```c
+int mesh_apply_connection_config_audio(MeshConnection *conn,
+                                       MeshConfig_Audio *cfg)
+```
+Applies the configuration to setup the connection payload for Audio packets
```
skolelis, line 112: "." at the end of sentence.

docs/sdk/SDK_API_DEFINITION.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+## mesh_err2str()
+```c
+const char *mesh_err2str(int err)
+```
+Gets a text description of the error code.
+
+#### Parameters
+* `[IN]` `err` – Error code returned from any Mesh DP API call.
```
skolelis, line 202: what is "DP"? "Data Plane" ? Define this abbreviation somewhere.

docs/sdk/SDK_API_WORKFLOW.md
```diff
@@ -0,0 +1,170 @@
...
+                                             .sample_rate
+                                             .format
+                                             .packet_time")
+```
+
+## Mesh DP SDK API internals – Connection configuration mindmap
+```mermaid
+mindmap
+)__MeshConnection__ configuration(
+   **.kind**
```
skolelis, line 130: ".kind" branch is drawn in yellow. (It is ok.) And it overwrites violet ".conn" branch. Maybe there is some way to enforce lack of overwriting?

ko80: Unfortunately, there is no option to control the rendering process of Mermaid.

ko80: Maybe I'll remove the mind map from this document since it doesn't bring more clarity to the configuration.

ffmpeg-plugin/mcm_audio_tx.c
```diff
@@ -28,33 +28,21 @@ typedef struct McmAudioMuxerContext {
     int sample_rate;
     char* ptime;
 
-    mcm_conn_context *tx_handle;
-    mcm_buffer *unsent_mcm_buf;
+    MeshClient *mc;
+    MeshConnection *conn;
+    MeshBuffer *unsent_buf;
     int unsent_len;
+
```
skolelis, line 35: empty line not needed

ffmpeg-plugin/mcm_common.c
```diff
@@ -4,156 +4,217 @@
...
+        if (err)
+            return err;
     } else {
-        av_log(avctx, AV_LOG_ERROR, "Unknown payload type\n");
-        return AVERROR(EINVAL);
+        MeshConfig_ST2110 cfg;
+
+        strlcpy(cfg.remote_ip_addr, ip_addr, sizeof(cfg.remote_ip_addr));
+        
+        if (kind == MESH_CONN_KIND_SENDER)
```
skolelis, line 100: I would add { / } else {  / }.

ffmpeg-plugin/mcm_common.c
```diff
@@ -4,156 +4,217 @@
...
-        param->payload_type = PAYLOAD_TYPE_ST22_VIDEO;
-        param->payload_codec = PAYLOAD_CODEC_JPEGXS;
-    } else if (!strcmp(payload_type, "st30")) {
-        param->payload_type = PAYLOAD_TYPE_ST30_AUDIO;
-    } else if (!strcmp(payload_type, "st40")) {
-        param->payload_type = PAYLOAD_TYPE_ST40_ANCILLARY;
-    } else if (!strcmp(payload_type, "rtsp")) {
-        param->payload_type = PAYLOAD_TYPE_RTSP_VIDEO;
+        err = mesh_apply_connection_config_memif(conn, &cfg);
+        if (err)
```
skolelis, line 93: this could be removed if at the end of function "return err" is done.

ffmpeg-plugin/mcm_common.c
```diff
@@ -4,156 +4,217 @@
...
+        /* transport type */
+        if (!strcmp(payload_type, "st20")) {
+            cfg.transport = MESH_CONN_TRANSPORT_ST2110_20;
+        } else if (!strcmp(payload_type, "st22")) {
+            cfg.transport = MESH_CONN_TRANSPORT_ST2110_22;
+        } else if (!strcmp(payload_type, "st30")) {
+            cfg.transport = MESH_CONN_TRANSPORT_ST2110_30;
+        } else {
+            av_log(avctx, AV_LOG_ERROR, "Unknown payload type\n");
+            return -MESH_ERR_CONN_CONFIG_INVAL;
```
skolelis, line 114: this "return" should be remained here (even if introduced other changes with return's).

ffmpeg-plugin/mcm_common.c
ko80: Assume the two options:
* Option 1
```
        if (kind == MESH_CONN_KIND_SENDER)
            cfg.remote_port = atoi(port);
        else
            cfg.local_port = atoi(port);
```
* Option 2
```
        if (kind == MESH_CONN_KIND_SENDER) {
            cfg.remote_port = atoi(port);
        } else {
            cfg.local_port = atoi(port);
        }
```
Both are valid, but Option 1 is typically used when both positive and negative branches contain only one line of code. The curvy brackets are redundant here. Option 2 is mostly for cases when you have more than one line of code in the block. I'd leave this one as it is.

ffmpeg-plugin/mcm_common.c
```diff
@@ -4,156 +4,217 @@
...
+        }
+
+        err = mesh_apply_connection_config_st2110(conn, &cfg);
+        if (err)
+            return err;
     }
 
     return 0;
```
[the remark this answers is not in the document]
ko80: Agree, what's you're proposing is correct. But having this `return` statement in this line clearly says that the function flow ends here. I suggest not to optimize the code prematurely.

ffmpeg-plugin/mcm_common.c
```diff
@@ -4,156 +4,217 @@
...
+            cfg.transport = MESH_CONN_TRANSPORT_ST2110_22;
+        } else if (!strcmp(payload_type, "st30")) {
+            cfg.transport = MESH_CONN_TRANSPORT_ST2110_30;
+        } else {
+            av_log(avctx, AV_LOG_ERROR, "Unknown payload type\n");
+            return -MESH_ERR_CONN_CONFIG_INVAL;
+        }
+
+        err = mesh_apply_connection_config_st2110(conn, &cfg);
+        if (err)
```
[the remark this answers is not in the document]
ko80: See my comment above https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/213#discussion_r1799457946.

ffmpeg-plugin/mcm_common.c
ko80: See my comment above https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/213#discussion_r1799457946

ffmpeg-plugin/mcm_video_rx.c
```diff
@@ -146,47 +117,65 @@ static int mcm_video_read_header(AVFormatContext* avctx)
...
 
 static int mcm_video_read_packet(AVFormatContext* avctx, AVPacket* pkt)
 {
     McmVideoDemuxerContext *s = avctx->priv_data;
-    mcm_buffer *buf = NULL;
-    int timeout = s->first_frame ? -1 : 1000;
-    int err = 0;
-    int ret;
+    MeshBuffer *buf;
+    int timeout = s->first_frame ? MESH_TIMEOUT_INFINITE : 1000;
```
skolelis, line 133: maybe it is worthy to pre-define this value 1000 ?
It is used 2 times.

ffmpeg-plugin/mcm_video_tx.c
```diff
@@ -29,85 +28,102 @@ typedef struct McmVideoMuxerContext {
     enum AVPixelFormat pixel_format;
     AVRational frame_rate;
 
-    mcm_conn_context *tx_handle;
-} McmVideoMuxerContext;
+    MeshClient *mc;
+    MeshConnection *conn;
+
+ } McmVideoMuxerContext;
```
skolelis, line 34: should teher be whitespace before "}" ?

ffmpeg-plugin/mcm_audio_rx.c
```diff
@@ -101,46 +129,57 @@ static int mcm_audio_read_header(AVFormatContext* avctx, enum AVCodecID codec_id
...
+    return mcm_audio_read_header(avctx, AV_CODEC_ID_PCM_S24BE);
 }
 
 static int mcm_audio_read_packet(AVFormatContext* avctx, AVPacket* pkt)
 {
     McmAudioDemuxerContext *s = avctx->priv_data;
-    mcm_buffer *buf = NULL;
-    int timeout = s->first_frame ? -1 : 1000;
-    int ret, len, err = 0;
+    int timeout = s->first_frame ? MESH_TIMEOUT_INFINITE : 1000;
```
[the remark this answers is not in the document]
ko80: The files `mcm_audio_rx.c` and `mcm_audio_tx.c` define two independent FFmpeg devices. Looking from each particular device perspective, this constant is used only once. Thus, there is no actual value of replacing it with a macro. I'd leave it as it is for now. Do not optimize prematurely.

ffmpeg-plugin/mcm_video_tx.c
ko80: Having a blank line here is not prohibited by the C standard. Sometimes I add blank lines to make the structure definition look better in the editor. So, this is to make the code easier to read.

ffmpeg-plugin/mcm_video_tx.c
```diff
@@ -117,8 +133,17 @@ static int mcm_video_write_packet(AVFormatContext* avctx, AVPacket* pkt)
...
+    if (err)
+        av_log(avctx, AV_LOG_ERROR, "Delete mesh connection failed: %s (%d)\n",
+               mesh_err2str(err), err);
+
+    err = mcm_put_client(&s->mc);
+    if (err)
+        av_log(avctx, AV_LOG_ERROR, "Put mesh client failed (%d)\n", err);
 
-    mcm_destroy_connection(s->tx_handle);
     return 0;
```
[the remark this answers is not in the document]
ko80: Yeah, it doesn't make sense to return any error here since it's the `write_trailer()` function which is called once the media stream transmission ends. After that, FFmpeg exits. So, this return value doesn't affect the flow.

docs/sdk/SDK_API_WORKFLOW.md
skolelis: it is quite clear. I just suggested to do it even more clear. Keep it that way.

ffmpeg-plugin/mcm_common.c
skolelis: not important. I commented removal of other "returns" in this function (to replace them with 1 at the end. But here there should be no change (even then).

ffmpeg-plugin/mcm_video_tx.c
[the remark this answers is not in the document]
skolelis: ok. so maybe change of declaration to "void function_name()" ?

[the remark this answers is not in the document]
ko80: Well, this function signature is predefined in FFmpeg. I can't change it :)
```
typedef struct FFOutputFormat {
...
    int (*write_header)(AVFormatContext *);
    /**
     * Write a packet. If FF_OFMT_FLAG_ALLOW_FLUSH is set in flags_internal,
     * pkt can be NULL in order to flush data buffered in the muxer.
     * When flushing, return 0 if there still is more data to flush,
     * or 1 if everything was flushed and there is no more buffered
     * data.
     */
    int (*write_packet)(AVFormatContext *, AVPacket *pkt);
    int (*write_trailer)(AVFormatContext *);
...
```

sdk/src/mesh_dp.c
```diff
@@ -0,0 +1,779 @@
...
+
+    conn_num = atomic_load_explicit(&mc_ctx->conn_num, memory_order_relaxed);
+    while (conn_num < mc_ctx->config.max_conn_num) {
+        if (atomic_compare_exchange_weak_explicit(&mc_ctx->conn_num, &conn_num,
+                                                  conn_num + 1,
+                                                  memory_order_release,
+                                                  memory_order_relaxed))
+            break;
+    }
+    if (!(conn_num < mc_ctx->config.max_conn_num))
```
Sakoram, line 156:
```suggestion
    if ((conn_num > mc_ctx->config.max_conn_num))
```
I think it is alright when `conn_num  == mc_ctx->config.max_conn_num`. We just don't want to have more then max.

tszumski:
Ready to merge?

ko80:
> Ready to merge?

I'm still working on resolving issues.

sdk/include/mesh_dp.h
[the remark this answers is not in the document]
ko80: This value order is no longer important since the maintainers agreed on switching to a JSON-based configuration in the next release.

sdk/include/mesh_dp.h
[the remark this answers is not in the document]
ko80: This discussion doesn't make much sense now because the maintainers agreed on switching to a JSON-based configuration in the next release.

tests/unit/mesh_dp_api_tests.cc
ko80: Will integrate that in the next unit test extension cycle.

sdk/include/mesh_dp.h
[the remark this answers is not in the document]
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

sdk/include/mesh_dp_internal.h
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

sdk/src/mesh_dp.c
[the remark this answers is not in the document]
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

sdk/src/mesh_dp.c
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.
@Sakoram please review.

docs/sdk/SDK_API_EXAMPLES.md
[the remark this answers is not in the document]
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_SMPTE_ST2110_RX_FLOW.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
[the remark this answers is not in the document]
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

docs/sdk/SDK_API_DEFINITION.md
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

ffmpeg-plugin/mcm_audio_tx.c
ko80: Fixed in eae8b083e0210567994eb379b5e228742451c44b.

ko80:
Unit tests are passing
```
[==========] Running 30 tests from 4 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from APITests_MeshClient
[ RUN      ] APITests_MeshClient.Test_CreateDeleteClient
[       OK ] APITests_MeshClient.Test_CreateDeleteClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_CreateClient
[       OK ] APITests_MeshClient.TestNegative_CreateClient (0 ms)
[ RUN      ] APITests_MeshClient.TestNegative_DeleteClient
[       OK ] APITests_MeshClient.TestNegative_DeleteClient (0 ms)
[----------] 3 tests from APITests_MeshClient (0 ms total)

[----------] 19 tests from APITests_MeshConnection
[ RUN      ] APITests_MeshConnection.TestNegative_DeleteClientLiveConnections
[       OK ] APITests_MeshConnection.TestNegative_DeleteClientLiveConnections (0 ms)
[ RUN      ] APITests_MeshConnection.Test_CreateEstablishShutdownDeleteConnection
[       OK ] APITests_MeshConnection.Test_CreateEstablishShutdownDeleteConnection (100 ms)
[ RUN      ] APITests_MeshConnection.Test_ApplyConfig
[       OK ] APITests_MeshConnection.Test_ApplyConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ApplyConfig
[       OK ] APITests_MeshConnection.TestNegative_ApplyConfig (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseConnectionConfig
[       OK ] APITests_MeshConnection.Test_ParseConnectionConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParseConnectionConfigInval
[       OK ] APITests_MeshConnection.TestNegative_ParseConnectionConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParseConnectionConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_ParseConnectionConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseVideoPayloadConfig
[       OK ] APITests_MeshConnection.Test_ParseVideoPayloadConfig (0 ms)
[ RUN      ] APITests_MeshConnection.Test_ParseAudioPayloadConfig
[       OK ] APITests_MeshConnection.Test_ParseAudioPayloadConfig (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParsePayloadConfigInval
[       OK ] APITests_MeshConnection.TestNegative_ParsePayloadConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ParsePayloadConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_ParsePayloadConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_EstablishConnectionConfigInval
[       OK ] APITests_MeshConnection.TestNegative_EstablishConnectionConfigInval (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnectionConfigIncompat
[       OK ] APITests_MeshConnection.TestNegative_CreateConnectionConfigIncompat (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_NulledClientAndConn (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_NulledConn (0 ms)
[ RUN      ] APITests_MeshConnection.Test_CreateConnection_MaxConnNumber
[       OK ] APITests_MeshConnection.Test_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_CreateConnection_MaxConnNumber
[       OK ] APITests_MeshConnection.TestNegative_CreateConnection_MaxConnNumber (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_ShutdownConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_ShutdownConnection_NulledConn (0 ms)
[ RUN      ] APITests_MeshConnection.TestNegative_DeleteConnection_NulledConn
[       OK ] APITests_MeshConnection.TestNegative_DeleteConnection_NulledConn (0 ms)
[----------] 19 tests from APITests_MeshConnection (100 ms total)

[----------] 7 tests from APITests_MeshBuffer
[ RUN      ] APITests_MeshBuffer.Test_GetPutBuffer
[       OK ] APITests_MeshBuffer.Test_GetPutBuffer (50 ms)
[ RUN      ] APITests_MeshBuffer.Test_GetBufferConnClosed
[       OK ] APITests_MeshBuffer.Test_GetBufferConnClosed (50 ms)
[ RUN      ] APITests_MeshBuffer.Test_GetBufferDefaultTimeout
[       OK ] APITests_MeshBuffer.Test_GetBufferDefaultTimeout (100 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf
[       OK ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledConnAndBuf (0 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledBuf
[       OK ] APITests_MeshBuffer.TestNegative_GetBuffer_NulledBuf (50 ms)
[ RUN      ] APITests_MeshBuffer.TestNegative_PutBuffer_NulledBuf
[       OK ] APITests_MeshBuffer.TestNegative_PutBuffer_NulledBuf (0 ms)
[ RUN      ] APITests_MeshBuffer.Test_ImportantConstants
[       OK ] APITests_MeshBuffer.Test_ImportantConstants (0 ms)
[----------] 7 tests from APITests_MeshBuffer (250 ms total)

[----------] 1 test from ProxyContextTests
[ RUN      ] ProxyContextTests.ProxyContextConstructor
[       OK ] ProxyContextTests.ProxyContextConstructor (0 ms)
[----------] 1 test from ProxyContextTests (0 ms total)

[----------] Global test environment tear-down
[==========] 30 tests from 4 test suites ran. (351 ms total)
[  PASSED  ] 30 tests.
```

tszumski (dismissed):
LGTM

Mionsz (dismissed):

sdk/src/mcm_dp.c
```diff
@@ -146,6 +146,10 @@ static void parse_memif_param(mcm_conn_param* request, memif_socket_args_t* memi
             log_error("Invalid audio parameters.");
         break;
 
+    case PAYLOAD_TYPE_RDMA_VIDEO:
+        request->payload_args.rdma_args.transfer_size = (size_t)request->payload_args.video_args.width * request->payload_args.video_args.height * 4;
```
Sakoram, line 150: I don't understand what is going on here. I believe, `parse_memif_param()` is called only when pure memif connection is created between apps, but we need to calculate transfer_size when RDMA connection is set up via media_proxy.

We could calculate it in media proxy itself, but probably better idea would be to calculate it before calling `mcm_create_connection_proxy()` 

I would also encourage you to calculate transfer_size with consideration of the video format, rather than just multiplying the number of pixels by 4.

Sakoram (approved):

PR #214, opened 2024-10-01
Do not exchange IP addresses between endpoints

Sakoram:
Previously, the endpoint waited to connect to another to obtain its IP address. The IP address is not needed because all endpoints are unidirectional. ( RX does not need TX's IP )

This fixes problems with creation of multiple sessions on single media_proxy.


ko80 (approved):
Looks good

PR #217, opened 2024-10-03
Wait for shared mem before receiving RDMA buffer

Sakoram:
There was a possibility that rx_rdma_consume_frame() was called from rx_rdma_frame_thread() before shared memory was ready. This caused timeout on memif_buffer_alloc_timeout() and frame was dropped.

### DIFF
--- a/media-proxy/src/rdma_session.c
+++ b/media-proxy/src/rdma_session.c
@@ -75,7 +75,7 @@ int rx_rdma_shm_init(rx_rdma_session_context_t *rx_ctx, memif_ops_t *memif_ops)
 
     /* Create memif interfaces
      * Both interaces are assigned the same socket and same id to create a loopback. */
-    rx_ctx->shm_ready = 0;
+    rx_ctx->shm_ready = ATOMIC_VAR_INIT(false);
     rx_ctx->memif_conn_args.socket = rx_ctx->memif_socket;
     rx_ctx->memif_conn_args.interface_id = memif_ops->interface_id;
     rx_ctx->memif_conn_args.buffer_size = (uint32_t)rx_ctx->transfer_size;
@@ -158,7 +158,7 @@ int tx_rdma_shm_init(tx_rdma_session_context_t *tx_ctx, memif_ops_t *memif_ops)
 
     /* Create memif interfaces
      * Both interaces are assigned the same socket and same id to create a loopback. */
-    tx_ctx->shm_ready = 0;
+    tx_ctx->shm_ready = ATOMIC_VAR_INIT(false);
     tx_ctx->memif_conn_args.socket = tx_ctx->memif_socket;
     tx_ctx->memif_conn_args.interface_id = memif_ops->interface_id;
     tx_ctx->memif_conn_args.buffer_size = (uint32_t)tx_ctx->transfer_size;
@@ -314,7 +314,7 @@ static void rx_rdma_consume_frame(rx_rdma_session_context_t *s, char *frame)
     uint32_t buf_size = s->transfer_size;
     uint16_t rx_buf_num = 0, rx = 0;
 
-    if (!s->shm_ready) {
+    if (!atomic_load(&s->shm_ready)) {
         INFO("%s memif not ready\n", __func__);
         return;
     }
@@ -344,6 +344,11 @@ static void *rx_rdma_frame_thread(void *arg)
     rx_rdma_session_context_t *s_ctx = (rx_rdma_session_context_t *)arg;
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
+
+    while (!atomic_load(&s_ctx->shm_ready) && !s_ctx->stop) {
+        INFO("%s memif not ready\n", __func__);
+        usleep(1000);
+    }
 
     printf("%s(%d), start\n", __func__, s_ctx->idx);
     while (!s_ctx->stop) {
--- a/media-proxy/src/rdma_session.c
+++ b/media-proxy/src/rdma_session.c
@@ -314,11 +314,6 @@ static void rx_rdma_consume_frame(rx_rdma_session_context_t *s, char *frame)
     uint32_t buf_size = s->transfer_size;
     uint16_t rx_buf_num = 0, rx = 0;
 
-    if (!atomic_load(&s->shm_ready)) {
-        INFO("%s memif not ready\n", __func__);
-        return;
-    }
-
     rx_bufs = s->shm_bufs;
 
     /* allocate memory */
@@ -345,14 +340,15 @@ static void *rx_rdma_frame_thread(void *arg)
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
 
-    while (!atomic_load(&s_ctx->shm_ready) && !s_ctx->stop) {
-        INFO("%s memif not ready\n", __func__);
+    while (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_acquire) && !s_ctx->stop)
         usleep(1000);
-    }
 
-    printf("%s(%d), start\n", __func__, s_ctx->idx);
+    printf("%s(%d), RX RDMA thread started\n", __func__, s_ctx->idx);
     while (!s_ctx->stop) {
         ep_recv_buf(cp_ctx, cp_ctx->data_buf, s_ctx->transfer_size);
+        if (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_acquire))
+            continue;
+
         rx_rdma_consume_frame(s_ctx, cp_ctx->data_buf);
     }
 
--- a/media-proxy/src/shm_memif_rdma.c
+++ b/media-proxy/src/shm_memif_rdma.c
@@ -38,7 +38,7 @@ int rx_rdma_on_connect(memif_conn_handle_t conn, void *priv_data)
 
     print_memif_details(conn);
 
-    rx_ctx->shm_ready = 1;
+    atomic_store(&rx_ctx->shm_ready, true);
 
     return 0;
 }
@@ -57,10 +57,10 @@ int rx_rdma_on_disconnect(memif_conn_handle_t conn, void *priv_data)
     }
 
     // release session
-    if (!rx_ctx->shm_ready) {
+    if (!atomic_load(&rx_ctx->shm_ready)) {
         return 0;
     }
-    rx_ctx->shm_ready = 0;
+    atomic_store(&rx_ctx->shm_ready, false);
 
     /* stop event polling thread */
     INFO("RX RDMA Stop poll event");
@@ -91,7 +91,7 @@ int tx_rdma_on_connect(memif_conn_handle_t conn, void *priv_data)
         return err;
     }
 
-    tx_ctx->shm_ready = 1;
+    atomic_store(&tx_ctx->shm_ready, true);
 
     print_memif_details(conn);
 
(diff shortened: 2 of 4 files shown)


media-proxy/src/rdma_session.c
```diff
@@ -345,6 +345,11 @@ static void *rx_rdma_frame_thread(void *arg)
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
 
+    while (!atomic_load(&s_ctx->shm_ready)) {
```
ko80, line 348: If the user or the system decides to shutdown the app, how do you exit this loop gracefully?

Sakoram: currently [when kill signal is received exit(0) is called ](https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/main/media-proxy/src/api_server_tcp.cc#L270) which terminates all threads.

It is not the most elegant way, I agree, but I don't see any possible critical outcomes.

ko80: Not handling graceful shutdown brings the following potential issues:
* Memory leaks
* System resources are not released (including networking sockets).
* 3rd party software components' resources are not released.
For now, I'm okay with using the existing approach of relying on `exit(1)`. But this must be reworked in the future to support graceful shutdown.

Please leave a comment here, something like `TODO: Implement graceful shutdown.`

Sakoram: Ok. I agree that we can leave this comment. This will be a good improvement, but I'm not sure if any memory leak or unreleased resources are possible. I think OS performs cleanup after a process is terminated and it release all resources that the process acquired.

Sakoram: Comment is added.

Sakoram: You know what. I removed the comment and added `&& !s_ctx->stop` to the condition. This is our graceful shutdown. We can stop this thread by setting the `stop` variable. It does not acquire any resources so exiting the loops is enough.

Another thing is if  the `stop` variable is well synchronized, in my view it is not, but maybe it is something to fix in different PR.

media-proxy/src/rdma_session.c
```diff
@@ -314,7 +314,7 @@ static void rx_rdma_consume_frame(rx_rdma_session_context_t *s, char *frame)
     uint32_t buf_size = s->transfer_size;
     uint16_t rx_buf_num = 0, rx = 0;
 
-    if (!s->shm_ready) {
+    if (!atomic_load(&s->shm_ready)) {
         INFO("%s memif not ready\n", __func__);
         return;
     }
```
ko80, line 320: Move this check to the calling function.
```suggestion
```

media-proxy/src/rdma_session.c
```diff
@@ -345,6 +345,11 @@ static void *rx_rdma_frame_thread(void *arg)
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
 
+    while (!atomic_load(&s_ctx->shm_ready) && !s_ctx->stop) {
+        INFO("%s memif not ready\n", __func__);
+        usleep(1000);
+    }
+
     printf("%s(%d), start\n", __func__, s_ctx->idx);
     while (!s_ctx->stop) {
```
ko80, line 354:
```suggestion
    while (atomic_load_explicit(&s_ctx->shm_ready, memory_order_relaxed) && !s_ctx->stop) {
```

Sakoram: looks better but it can take some time between this check and calling rx_rdma_consume_frame() as ep_recv_buf() publishes buffer to receive incoming data into it and waits till transmission is complete. 
So, potentially we can call memif_buffer_alloc() on non existing connection which probably will just return error and nothing tragic would happen, but still, I would prefer to have a check closer to the call to memif_buffer_alloc()

media-proxy/src/rdma_session.c
```diff
@@ -345,6 +345,11 @@ static void *rx_rdma_frame_thread(void *arg)
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
 
+    while (!atomic_load(&s_ctx->shm_ready) && !s_ctx->stop) {
```
ko80, line 348:
```suggestion
    while (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_acquire) && !s_ctx->stop) {
```

media-proxy/src/rdma_session.c
```diff
@@ -345,6 +345,11 @@ static void *rx_rdma_frame_thread(void *arg)
...
+        INFO("%s memif not ready\n", __func__);
+        usleep(1000);
+    }
+
     printf("%s(%d), start\n", __func__, s_ctx->idx);
     while (!s_ctx->stop) {
         ep_recv_buf(cp_ctx, cp_ctx->data_buf, s_ctx->transfer_size);
```
ko80, line 355:
```suggestion
        ep_recv_buf(cp_ctx, cp_ctx->data_buf, s_ctx->transfer_size);
        if (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_relaxed)
            continue;
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -38,7 +38,7 @@ int rx_rdma_on_connect(memif_conn_handle_t conn, void *priv_data)
 
     print_memif_details(conn);
 
-    rx_ctx->shm_ready = 1;
+    atomic_store(&rx_ctx->shm_ready, true);
```
ko80, line 41:
```suggestion
    atomic_store_explicit(&rx_ctx->shm_ready, true, memory_order_release);
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -57,10 +57,10 @@ int rx_rdma_on_disconnect(memif_conn_handle_t conn, void *priv_data)
     }
 
     // release session
-    if (!rx_ctx->shm_ready) {
+    if (!atomic_load(&rx_ctx->shm_ready)) {
         return 0;
     }
-    rx_ctx->shm_ready = 0;
+    atomic_store(&rx_ctx->shm_ready, false);
```
ko80, line 63:
```suggestion
    atomic_store_explicit(&rx_ctx->shm_ready, false, memory_order_release);
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -57,10 +57,10 @@ int rx_rdma_on_disconnect(memif_conn_handle_t conn, void *priv_data)
     }
 
     // release session
-    if (!rx_ctx->shm_ready) {
+    if (!atomic_load(&rx_ctx->shm_ready)) {
```
ko80, line 60:
```suggestion
    if (!atomic_load_relaxed(&rx_ctx->shm_ready, memory_order_relaxed)) {
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -91,7 +91,7 @@ int tx_rdma_on_connect(memif_conn_handle_t conn, void *priv_data)
         return err;
     }
 
-    tx_ctx->shm_ready = 1;
+    atomic_store(&tx_ctx->shm_ready, true);
```
ko80, line 94:
```suggestion
    atomic_store_explicit(&tx_ctx->shm_ready, true, memory_order_release);
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -113,10 +113,10 @@ int tx_rdma_on_disconnect(memif_conn_handle_t conn, void *priv_data)
     }
 
     // release session
-    if (!tx_ctx->shm_ready) {
+    if (!atomic_load(&tx_ctx->shm_ready)) {
```
ko80, line 116:
```suggestion
    if (!atomic_load_explicit(&tx_ctx->shm_ready, memory_order_relaxed)) {
```

media-proxy/src/shm_memif_rdma.c
```diff
@@ -113,10 +113,10 @@ int tx_rdma_on_disconnect(memif_conn_handle_t conn, void *priv_data)
...
         return 0;
     }
-    tx_ctx->shm_ready = 0;
+    atomic_store(&tx_ctx->shm_ready, false);
```
ko80, line 119:
```suggestion
    atomic_store_explicit(&tx_ctx->shm_ready, false, memory_order_release);
```

media-proxy/src/rdma_session.c
Sakoram: I used `memory_order_acquire` instead of `memory_order_relaxed` because relaxed is ok when the state changes to not ready and we want to skip rx_rdma_consume_frame(), but then if shmem is again ready we need `memory_order_acquire`

media-proxy/src/rdma_session.c
Sakoram: done

media-proxy/src/shm_memif_rdma.c
Sakoram: done

media-proxy/src/shm_memif_rdma.c
Sakoram: done

media-proxy/src/shm_memif_rdma.c
Sakoram: Here I used relaxed instead of release. We don't modify any memory before this store operation, so there is no need to barrier.

media-proxy/src/shm_memif_rdma.c
Sakoram: done

media-proxy/src/shm_memif_rdma.c
Sakoram: done

media-proxy/src/shm_memif_rdma.c
Sakoram: Again. Here I used relaxed instead of release. We don't modify any memory before this store operation, so there is no need to barrier.

media-proxy/src/rdma_session.c
```diff
@@ -345,9 +340,19 @@ static void *rx_rdma_frame_thread(void *arg)
     libfabric_ctx *rdma_ctx = s_ctx->rdma_ctx;
     ep_ctx_t *cp_ctx = s_ctx->ep_ctx;
 
-    printf("%s(%d), start\n", __func__, s_ctx->idx);
+    while (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_acquire) && !s_ctx->stop) {
+        DEBUG("%s memif not ready\n", __func__);
```
ko80, line 344: I suggest to avoid printing that every 1ms.
```suggestion
```

media-proxy/src/rdma_session.c
```diff
@@ -345,9 +340,19 @@ static void *rx_rdma_frame_thread(void *arg)
...
+        usleep(1000);
+    }
+
+    printf("%s(%d), RX RDMA thread started\n", __func__, s_ctx->idx);
     while (!s_ctx->stop) {
+
         ep_recv_buf(cp_ctx, cp_ctx->data_buf, s_ctx->transfer_size);
+        if (!atomic_load_explicit(&s_ctx->shm_ready, memory_order_acquire)) {
+            DEBUG("memif disconnected");
```
ko80, line 353: I suggest to avoid printing that constantly in the loop.
```suggestion
```

Sakoram: ok, I will remove it entirely as a log is printed in place where the variable is changed.

media-proxy/src/rdma_session.c
Sakoram: ok, I will remove it entirely as a log is printed in places where the variable is changed.

ko80 (dismissed):
Looks very good. Thank you Marek.

ko80 (approved):

PR #220, opened 2024-10-07
Fix: resolve execution failure in test.sh script

Mionsz:
Fix: resolve execution failure in test.sh script
Remove quotes form run in background method

### DIFF
--- a/tests/single-node-sample-apps/test.sh
+++ b/tests/single-node-sample-apps/test.sh
@@ -81,7 +81,7 @@ function run_in_background() {
     # Returns PID of a spawned process
     info "Run: ${1}"
     echo -e "${1}\n" > "${2}"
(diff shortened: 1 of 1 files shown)


Sakoram (approved):

DawidWesierski4:
> Quick Fix - update test.sh. Remove quotes from run in background method.

is the Quick Fix necessary ?

to adhere to the commit message conventions could be better to call it 

```
Fix: fix test.sh execution failing
Remove quotes form run in background method
```
or something of similar nature would  be a ittle bit more readable as a commit message


But lgtm

Sakoram (changes requested):
Change commit message format according to Dawid's comment

MateuszGrabuszynski (approved):
lgtm

PR #227, opened 2024-10-09
Update scripts and docs

Sakoram:


Mionsz (approved):
Merge it when you are ready - or in case any changes I will make to my PR, DM me.

scripts/setup_build_env.sh
```diff
@@ -124,14 +123,17 @@ function install_ubuntu_package_dependencies()
...
         autoconf \
         automake \
         autotools-dev \
         libtool \
         grpc-proto \
         protobuf-compiler-grpc \
-        libgrpc10
+        libgrpc10 \
+        dracut \
+        linux-headers-"$(uname -r)"
```
Mionsz, line 136: Ok, maybe this one could be improved as you do not collect return code from shell execution (probably do not need to most of the time) and this triggers shell-check warnings. Can we have it done like:
```bash
{
local APT_LINUX_HEADERS="linux-headers-$(uname -r)"
# Install package dependencies
apt-get update --fix-missing && \
apt-get install --no-install-recommends -y \
       ...
       "${APT_LINUX_HEADERS}"
}
```

PR #232, opened 2024-10-11
fix trivy scans

zLukas:
Changelog:
* Trivy set to scan actual repo not empty workspace
* upload scan results to github security tab
* Deployment (with helmcharts) ignored during scan as it's not supported yet


soopel (approved):
Looks good to me

PR #234, opened 2024-10-16
gRPC light removal

tszumski:
Remove untested/broken gRPC code


ko80 (approved):
Looks good.

Before merging, make sure this change doesn't brake a sample video or audio transmission between two FFmpeg plugins.

tszumski:
I can confirm audio/video streams works as expected

Sakoram (approved):
The PR looks good. but...
I've noticed that `ProxyContext::Stop()` is the only place where functions like
- st_pthread_mutex_destroy(&sessions_count_mutex_lock);
- mtl_deinit(mDevHandle);
- rdma_deinit(&mDevHandle_rdma);

are called. We should create some mechanism that will clean all those resources. Currently OS does it, but I think the application itself should do it.

PR #235, opened 2024-10-17
Remove udp/rtsp h264

tszumski:
Remove udp/rtsp h264, 
To be merged after: https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/213


media-proxy/include/proxy_context.h
```diff
@@ -44,8 +44,6 @@ class ProxyContext {
     libfabric_ctx *mDevHandle_rdma = NULL;
 
     /*udp pool*/
```
Sakoram, line 46: remove the comment too

media-proxy/src/proxy_context.cc
```diff
@@ -570,7 +570,7 @@ int ProxyContext::RxStart_mtl(const mcm_conn_param *request)
     int ret;
 
     /*add lock to protect MTL library initialization to aviod being called by multi-session simultaneously*/
-    if (!mDevHandle && imtl_init_preparing == false) {
+    if (mDevHandle == NULL && imtl_init_preparing == false) {
```
Sakoram, line 573: Why this change? We usually use !val to check if value is NULL.

tszumski: To align with code in TxStart_mtl()

ko80: > To align with code in TxStart_mtl()

Could you elaborate on that?

Sakoram: I guess we just have to define our approach. In a new code we tend to use `!val`, in old code there is a mix of everything.
Let's agree on something. Me personally I'm okay with both approaches. `val == NULL` is slightly more readable, `!val` is smaller and more elegant.

ko80: I vote for `if (!something) {}`.

tszumski: I also vote for ```!val```

Sakoram: I declare to all and sundry that the winner is `!val` :tada:

media-proxy/include/proxy_context.h
tszumski: done

media-proxy/src/proxy_context.cc
tszumski: Removed this change

Sakoram (approved):
LGTM, but before merging, please run our samples from tests/single-node-sample-apps/ to make sure everything is in ok

ko80 (approved):
LGTM

PR #236, opened 2024-10-17
zLukas/trivy-raporting upgrade

zLukas:
ADD:
* Trivy will fail if it find any errors, and by default, it stops the rest of the pipeline, set upload steps, upload SARIF and update security tab regardless, the Trivy scan status
* add second trivy status with table output (human redable)


soopel (dismissed):
ok to merge

soopel (approved):
Build process fixed. Ok to merge

PR #239, opened 2024-10-21
Refactor libfabric's CQ handling

Sakoram:
Refactor libfabric's CQ handling:
- create just one CQ per endpoint
- always read just one event
- set read method at init instead of checking it at every read

* This PR should fix coverity issue


media-proxy/src/libfabric_cq.c
```diff
@@ -78,41 +81,28 @@ void rdma_cq_set_wait_attr(struct fi_cq_attr *cq_attr, enum cq_comp_method metho
...
-        if (ret > 0) {
-            if (timeout >= 0)
-                clock_gettime(CLOCK_MONOTONIC, &a);
-            (*cur)++;
-        } else if (ret < 0 && ret != -FI_EAGAIN) {
+        ret = fi_cq_read(ep_ctx->cq_ctx.cq, entry, 1);
+        if (ret < 0 && ret != -FI_EAGAIN) {
             return ret;
-        } else if (timeout >= 0) {
+        } else if (ret == -FI_EAGAIN && timeout >= 0) {
```
ko80, line 96: Not sure about checking for `-FE_EAGAIN` twice here. Maybe slightly rework the flow to have just one check?

ko80 (approved):
Looks good. I like that you write the code which is easy to read. Well done!

media-proxy/src/libfabric_cq.c
Sakoram: I rewrote the function. Now it look better

media-proxy/src/libfabric_cq.c
```diff
@@ -78,43 +81,31 @@ void rdma_cq_set_wait_attr(struct fi_cq_attr *cq_attr, enum cq_comp_method metho
...
+            ep_ctx->cq_ctx.cq_cntr++;
+            return 0;
         } else if (ret < 0 && ret != -FI_EAGAIN) {
             return ret;
-        } else if (timeout >= 0) {
-            clock_gettime(CLOCK_MONOTONIC, &b);
-            if ((b.tv_sec - a.tv_sec) > timeout) {
+        }
+
+        if (timeout >= 0) {
```
ko80, line 101: This code will be running when `ret == 0` or `ret == -FI_EAGAIN`, correct?
If yes, please add a comment explaining that.

media-proxy/src/libfabric_cq.c
```diff
@@ -269,6 +198,45 @@ int rdma_cq_readerr(struct fid_cq *cq)
...
+    case RDMA_COMP_SREAD:
+    case RDMA_COMP_YIELD:
+        ep_ctx->cq_ctx.eq_read = rdma_wait_for_comp;
+        break;
+    case RDMA_COMP_WAIT_FD:
+        ep_ctx->cq_ctx.eq_read = rdma_fdwait_for_comp;
+        break;
+    default:
+        ep_ctx->cq_ctx.eq_read = rdma_spin_for_comp;
+        break;
```
ko80, line 234: This is redundant. But keep it if you'd like :)

ko80 (dismissed):

media-proxy/src/libfabric_cq.c
Sakoram: Correct. ret is of the ssizze_t type, it returns number of cq entries returned or an error.
So we are satisfied when ret == 1,  as we requesting just one entry, then we want to return 0, means success.
if any error other then -FI_EAGAIN occurred we want to immediately return it.
-FI_EAGAIN means that nothing is in the queue so we're spin waiting for something in queue ( and are checking if no time out).

I'm not sure if ret ever will be equal to 0, but anyway it does not mean success.

ko80: Having such a comment in the code would be very helpful for future efforts of development.

Sakoram: Do you think a comment in code is necessary? maybe I will change ret to ssizze_t  to better communicate  intention

Sakoram: comments added :)

ko80 (approved):

PR #240, opened 2024-10-23
Replace current c-style sessions to c++

tszumski:


media-proxy/include/session-base.h
```diff
@@ -0,0 +1,59 @@
...
+
+using namespace std; // for atomic_bool
+
+#include <mcm_dp.h>
+
+#include "shm_memif.h" /* share memory */
+
+#include "utils.h"
+
+class session
```
Sakoram, line 25: Let's capitalize class names

media-proxy/include/session-base.h
```diff
@@ -0,0 +1,59 @@
...
+#define __SESSION_BASE_H
+
+#include <limits.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <atomic>
+#include <thread>
+#include <stdexcept>
+
+using namespace std; // for atomic_bool
```
Sakoram, line 17: first you write `using namespace std;` and after that sometimes in code there is `std::` used. Let's unify approach. 

I would explicitly use `std::`, it makes obvious that given function is standard one.

media-proxy/include/session-base.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <mcm_dp.h>
+
+#include "shm_memif.h" /* share memory */
```
Sakoram, line 21:
```suggestion
#include "shm_memif.h" /* shared memory */
```

media-proxy/include/session-base.h
```diff
@@ -0,0 +1,59 @@
...
+    memif_conn_handle_t memif_conn;
+    memif_conn_args_t memif_conn_args;
+
+  private:
+    std::thread *memif_event_thread;
+    void memif_event_loop();
+    int shm_deinit();
+
+  protected:
+    atomic_bool shm_ready;
```
Sakoram, line 45: you split public section into fields and methods, but in protected section you have both fields and methods. It is a little inconsistent.

Honestly, in my very limited experience I rarely encountered splitting the sections, but I'm ok with it as far as it is consistent.

media-proxy/include/session-mtl.h
```diff
@@ -0,0 +1,214 @@
...
+#ifndef __SESSION_MTL_H
+#define __SESSION_MTL_H
+
+#include <mtl/st30_api.h>        /* st2110-30 */
+#include <mtl/st40_api.h>        /* st2110-40 */
+#include <mtl/st_pipeline_api.h> /* st2110 pipeline */
+
+#include "session-base.h"
+
+#define MTL_ZERO_COPY
```
Sakoram, line 16: This flag is added by build script, please remove it from the code.

media-proxy/src/api_server_tcp.cc
```diff
@@ -134,63 +136,9 @@ void* msg_loop(void* ptr)
                 if (it->id == session_id) {
                     /* return memif parameters. */
                     memif_conn_param param = { };
-                    if (it->type == TX) {
```
Sakoram, line 137: beautiful :tada: I was devastated every time I looked at that spaghetti. good job

media-proxy/src/api_server_tcp.cc
```diff
@@ -134,63 +136,9 @@ void* msg_loop(void* ptr)
                 if (it->id == session_id) {
```
Sakoram, line 136: please change `Failed to start MTL session.` to `Failed to start a session.`  a few lines above. ^^^^ 
Thanks. :)

media-proxy/include/session-mtl.h
tszumski: Previous approach was wrong. Macro is defined in media_proxy subdirectory.
When you add ```#include "session-mtl.h"``` inside Unit Tests macro is alwas OFF, while compilation of "media_proxy" submodule happen with macro ON. This lead to smaller object allocation size compared to what library expected, and overall read/write outsied of allocation happen

media-proxy/src/session-base.cc
```diff
@@ -0,0 +1,211 @@
...
+    session *connection = (session *)priv_data;
+    return connection->on_receive_cb(conn, qid);
+}
+
+void session::memif_event_loop()
+{
+    int err;
+    memif_socket_handle_t memif_socket = memif_conn_args.socket;
+
+    do {
```
Sakoram, line 42: I think we should implement graceful exit here. Currently we use pthread_cancel(), which could potentially create some problems as this thread handles events and calls functions like on_receive_cb(). Those functions lock mutexes, and interact with libfabric and MTL. A lot of funny staff can happen when this is interrupted in random places.

We could set a little shorter timeout then infinity and check some new `stop` flag on every loop.

media-proxy/include/session-mtl.h
tszumski: TODO:
Remove ZERO_COPY macro from media_proxy/CMakeLists.txt

media-proxy/include/session-base.h
tszumski: Do you mean CamelCase or Camel_Snake ?

media-proxy/include/session-base.h
tszumski: Agree, then this change also apply for atomic_bool

media-proxy/include/session-base.h
Sakoram: I don't want to change snake_case, but personally I'm used to class names starting with capital letters, so for classes it would be Snake_case_started_with_capital_letter (however it is called) ^^

media-proxy/include/session-base.h
Sakoram: yeah, I hope it is possible to do without breaking C code.

media-proxy/include/session-base.h
tszumski: I would suggest JustForClassNames, to be consistent with gRPC/TCP classes

media-proxy/include/session-base.h
tszumski: Yes, I can do that

media-proxy/src/session-base.cc
tszumski: Good idea!

media-proxy/src/session-base.cc
```diff
@@ -0,0 +1,211 @@
...
+
+    INFO("MEMIF DISCONNECTED.");
+}
+
+int session::shm_init(memif_ops_t *memif_ops, uint32_t buffer_size, uint32_t log2_ring_size)
+{
+    int ret = 0;
+
+    /* Set application name */
+    strncpy(memif_socket_args.app_name, memif_ops->app_name,
```
Sakoram, line 56:
```suggestion
    strlcpy(memif_socket_args.app_name, memif_ops->app_name,
```
to be sure that the string is NULL terminated. Should be done everywhere.

media-proxy/src/session-mtl-st2110-20.cc
```diff
@@ -0,0 +1,416 @@
...
+    if (!frame_thread_handle) {
+        cleanup();
+        throw runtime_error("Failed to create thread");
+    }
+}
+
+void rx_st20_mtl_session::cleanup()
+{
+    if (frame_thread_handle) {
+        frame_thread_handle->join();
```
Sakoram, line 108: I think it won't cancel the thread. It would wait indefinitely. The stop variable has to be set for this thread to stop.

Sakoram: The same issue exists in different places

Sakoram:
I'm thinking what to do with error handling.
- In some places there is simply `return -1` used. I don't like it.
- Elsewhere, some semi-useful error codes ( like `return -EINVAL` ) are returned. That's better, the best what can be done in C, maybe only defining more descriptive err codes could improve the situation in C
- But this is c++ and sometimes throw is used, which seems to be the most elegant way. maybe slightly less performant, but not much, I guess it is the best everywhere besides critical paths. 

Any thoughts? Are there any drawback of using try-catch and throw everywhere?

Sakoram:
What do you think about splitting every class to a separate file? It could improve readability

media-proxy/src/proxy_context.cc
```diff
@@ -538,32 +537,34 @@ int ProxyContext::RxStart_rdma(const mcm_conn_param *request)
...
+    session_ptr->payload_type = request->payload_type;
+    session_ptr->id = memif_ops.m_session_count;
+    INFO("%s, session id: %d", __func__, session_ptr->id);
+    session_ptr->type = RX;
+    mDpCtx.push_back(session_ptr);
+
+    return session_ptr->id;
 }
 
 int ProxyContext::RxStart_mtl(const mcm_conn_param *request)
```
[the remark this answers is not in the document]
Sakoram: The "case switch part" could be even a separate class as it is basically a factory pattern. Calling it explicitly a factory could help readability. 
(if we want to be very crazy, we could even use abstract factory with concrete factories for RX and TX. I guess this could also work, but this is a lot of abstraction)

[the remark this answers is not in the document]
tszumski: Any bigger changes to proxy_context.cc and api_server_tcp.cc should be done in the next PR to not to touch too many topics at once

media-proxy/src/proxy_context.cc
```diff
@@ -828,96 +763,22 @@ int ProxyContext::TxStart(const mcm_conn_param *request)
...
-            mtl_st20p_tx_session_stop((*ctx)->tx_session);
-            mtl_st20p_tx_session_destroy(&(*ctx)->tx_session);
-            break;
-        }
+        INFO("%s, Stop session ID: %d", __func__, session_id);
 
+        session *session_ptr = *ctx;
+        session_ptr->session_stop();
+        delete session_ptr;
         mDpCtx.erase(ctx);
```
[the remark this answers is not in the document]
tszumski: Any bigger changes to proxy_context.cc and api_server_tcp.cc should be done in the next PR to not to touch too many topics at once

media-proxy/src/session-mtl-st2110-20.cc
```diff
@@ -0,0 +1,416 @@
...
+    /* allocate memory */
+    uint16_t qid = 0;
+    uint16_t rx_buf_num = 0;
+
+    if (!atomic_load_explicit(&shm_ready, memory_order_relaxed)) {
+        ERROR("rx_st20p_query_ext_frame: MemIF connection not ready.");
+        return -1;
+    }
+
+    err = memif_buffer_alloc_timeout(memif_conn, qid, shm_bufs, 1, &rx_buf_num, frame_size, 10);
```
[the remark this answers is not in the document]
tszumski: Good catch

tszumski:
> I'm thinking what to do with error handling.
> 
> * In some places there is simply `return -1` used. I don't like it.
> * Elsewhere, some semi-useful error codes ( like `return -EINVAL` ) are returned. That's better, the best what can be done in C, maybe only defining more descriptive err codes could improve the situation in C
> * But this is c++ and sometimes throw is used, which seems to be the most elegant way. maybe slightly less performant, but not much, I guess it is the best everywhere besides critical paths.
> 
> Any thoughts? Are there any drawback of using try-catch and throw everywhere?

- We cannot use "throw" functionality in function that are called through callback by MTL/memif. We can only return error codes or print error messages
- Generally for proper error(and messages/logs) handling we have to implement one consistent approach across entire MCM repository. I think this is quite large topic and should added as another PR

tszumski:
> What do you think about splitting every class to a separate file? It could improve readability

Good idea, only if we move source files into multiple subdirectories. Otherwise having too many files in single directory will reduce readability
My proposition:
```
media_proxy
    app
        media_proxy.cc
    servers
        tcp
            <tcp-related sources>
        grpc
            <grpc-related sources>
    sessions
        rdma
            <rdma-related sources>
        mtl
            <mtl-related sources>
```

media-proxy/src/proxy_context.cc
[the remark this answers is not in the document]
tszumski: > * Break this function into a few smaller ones. for example move device init into first function, move the case switch to second, call the two of them and set variables in this one.
> * change imtl_init_preparing to some real lock
> * the same applies to TxStart_mtl
> * please add a lock similar to imtl_init_preparing (but actually working) also to RxStart_rdma and TxStart_rdma

I think the best way is to move MTL/RDMA device creation before we accept any connection(both of them), and close those devices after we stop accepting connection. This way we avoid unnecessary locks and imtl_init_preparing will disappear. Thanks to that only mDpCtx have to be locked

media-proxy/src/session-mtl-st2110-20.cc
tszumski: stop variable is set in session_stop()

media-proxy/src/proxy_context.cc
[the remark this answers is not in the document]
Sakoram: I think that we can define some arguments to media_proxy to specify what backend will be initialized at startup. maybe everything availably should be initialized by default.

media-proxy/src/session-mtl-st2110-20.cc
Sakoram: but session_stop() is called from proxy_context.cc, and this join() will freeze the thread that called rx_st20_mtl_session(), which is the thread from proxy_context.cc which would call session_stop() otherwise.

Anyway, you are right. We need to leverage RAII as much as possible, we should move to destructors every cleanup that can be moved there.

Sakoram: In this case, to cleanup when constructor fails and throws an exception we could:
1. Use smart pointers whenever you allocate memory
2. Do something like this when smart pointers aren't enough:
```
class Resource {
    // Your resource management logic
public:
    Resource() {
        // Acquire the resource
    }
    ~Resource() {
        // Release the resource
    }
};

class Session{
    Resource resource;  // RAII type
public:
    Session() {
        // If an exception is thrown here, `resource` will still be cleaned up
        if (/* some condition */) {
            throw std::runtime_error("Initialization failed");
        }
    }

    ~Session() {
        // `Resource` will automatically clean itself up
    }
};
```

media-proxy/include/session-rdma.h
```diff
@@ -0,0 +1,86 @@
...
+#ifndef __SESSION_RDMA_H
+#define __SESSION_RDMA_H
+
+#include <stdlib.h>
+#include <mcm_dp.h>
+
+#include "libfabric_dev.h"
+#include "libfabric_ep.h"
+#include "session-base.h"
+using namespace std;
```
ko80, line 16: Avoid using `std` this way. As @Sakoram suggested, switch to passing it explicitly, e.g. `std::mutex`.

media-proxy/include/session-rdma.h
```diff
@@ -0,0 +1,86 @@
...
+    enum direction dir;
+} rdma_s_ops_t;
+
+class tx_rdma_session : public session
+{
+  public:
+    libfabric_ctx *rdma_ctx;
+    ep_ctx_t *ep_ctx;
+
+    volatile bool stop;
```
ko80, line 36: The fact that you use `volatile` is really suspicious. I'm not insisting on reworking that immediately, but I strongly recommend to consider using an atomic variable.

media-proxy/src/api_server_tcp.cc
```diff
@@ -212,16 +160,8 @@ void* msg_loop(void* ptr)
...
-                if (it->id == session_id) {
-                    if (it->type == TX) {
-                        proxy_ctx->TxStop(session_id);
-                    } else {
-                        proxy_ctx->RxStop(session_id);
-                    }
-                    sessionKeepRunning = false;
-                    break;
-                }
+            if(!proxy_ctx->Stop(session_id)){
```
ko80, line 163: You hate spaces, don't you? ;-)
```suggestion
            if (!proxy_ctx->Stop(session_id)) {
```

media-proxy/src/session-base.cc
ko80: Agree. But when changing `strncpy` to `strlcpy` please note you don't need to subtract 1 from the size.

media-proxy/src/session-base.cc
```diff
@@ -0,0 +1,211 @@
...
+}
+
+session::session() : id(0), memif_socket(0), memif_conn(0)
+{
+    memif_event_thread = 0;
+    type = TX;
+    payload_type = PAYLOAD_TYPE_NONE;
+    memif_socket_args = {0};
+    memif_conn_args = {0};
+}
```
ko80, line 209: What's the reason to split initialization of the fields into the two parts, the function header and the function body?

media-proxy/src/api_server_tcp.cc
Sakoram: I recommend using: `git-clang-format main`

the `.clang-format` file is configured to format everything approximately according to our codestyle.

media-proxy/src/session-mtl-st2110-20.cc
```diff
@@ -0,0 +1,416 @@
...
+}
+
+/* Create new RX session */
+rx_st20_mtl_session::rx_st20_mtl_session(mtl_handle dev_handle, struct st20p_rx_ops *opts,
+                                         memif_ops_t *memif_ops)
+    : handle(0), frame_thread_handle(0), shm_bufs(0), fb_recv(0)
+{
+    if (!dev_handle || !opts || !memif_ops) {
+        cleanup();
+        throw invalid_argument("Input parameters cannot be null");
```
ko80, line 53: I have a strong suggestion not to throw exceptions but return errors on all layers of the calling stack and process return values explicitly.

media-proxy/src/api_server_tcp.cc
tszumski: I did it for all new files. Applying clang-format for existing files can show much more changes and reduce readability

media-proxy/src/session-mtl-st2110-20.cc
tszumski: Constructor cannot return error. The only way to handle error path nicely is to use throw

media-proxy/src/session-mtl-st2110-22.cc
```diff
@@ -0,0 +1,408 @@
...
+        throw runtime_error("Failed to create thread");
+    }
+}
+
+void rx_st22_mtl_session::cleanup()
+{
+    if (frame_thread_handle) {
+        frame_thread_handle->join();
+        delete frame_thread_handle;
+        frame_thread_handle = 0;
```
ko80, line 107: Maybe replace `0` with `NULL` or `nullptr`?

media-proxy/src/api_server_tcp.cc
Sakoram: you can use clang-format not only to format files but also to format commits. `git-clang-format main` formats everything changed in comparison to the `main` branch.

media-proxy/src/session-mtl-st2110-20.cc
ko80: You can extract the functionality that can fail in an `init()` method which returns an error.
The code could look as below
```c
A *a = new A;
int err = a->init();
if (err) {
    delete a;
    return err;
}
```
Simple. Easy to read.

ko80:
> I'm thinking what to do with error handling.
> 
> * In some places there is simply `return -1` used. I don't like it.
> * Elsewhere, some semi-useful error codes ( like `return -EINVAL` ) are returned. That's better, the best what can be done in C, maybe only defining more descriptive err codes could improve the situation in C
> * But this is c++ and sometimes throw is used, which seems to be the most elegant way. maybe slightly less performant, but not much, I guess it is the best everywhere besides critical paths.
> 
> Any thoughts? Are there any drawback of using try-catch and throw everywhere?

Consider returning error, not throwing exceptions.

media-proxy/src/session-mtl-st2110-20.cc
Sakoram: I totally agree that we have to handle errors in places where they occur, but I'm not that sure that doing that by return values is significantly better then handling exceptions.

If we chose to not throw in constructors, i would do something like this, and make the constructor private:
```c++
class A{
public:
    static std::optional<A> create(size_t size) {
        A obj;
        if (obj.initialize(size)) {
            return obj;  // Return constructed object
        }
        return std::nullopt;  // Indicate failure to construct
    }


private:
    A() = default;

    bool initialize(size_t size) {
        try {
            data = new int[size];  // May throw std::bad_alloc
            return true;
        } catch (const std::bad_alloc&) {
            std::cerr << "Memory allocation failed in initialize." << std::endl;
            return false;
        }
    }

    ~A() {
        delete[] data;
    }

    int* data = nullptr;
};

int main() {
    size_t size = 1000000;  // Large size for demonstration

    // Try to create the A using the factory method
    auto objOpt = A::create(size);

    if (objOpt) {
        // Object was successfully created, use it
    } else {
        // Handle the failure to create the object
        std::cerr << "Failed to create A due to memory allocation issues." << std::endl;
    }
```

ko80: Exactly. In this example you don't throw your own exceptions. However, you catch exceptions that can occur in a 3rd party component. And then you return an error. That's what I want to achieve.

I suggest we establish such a strategy of three statements:
* Don't throw own exceptions.
* Always catch 3rd party component exceptions.
* Always handle and return errors.

ko80: Another **huge** drawback of implementing some functionality in a constructor is that you can't call overridden virtual methods in the constructor. It means the base class can't call a virtual method implemented in the derived class. The only option is to only initialize the class instance in the constructor, and when the instance is created, call a virtual method to do what your feature should do.

media-proxy/include/session-base.h
tszumski: done

media-proxy/include/session-base.h
tszumski: done

media-proxy/include/session-base.h
tszumski: done

media-proxy/include/session-mtl.h
tszumski: done

media-proxy/src/api_server_tcp.cc
tszumski: done

media-proxy/src/session-base.cc
tszumski: done

media-proxy/src/session-mtl-st2110-20.cc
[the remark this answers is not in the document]
tszumski: done, _timeout is only used in non-callback path

media-proxy/include/session-rdma.h
tszumski: done

media-proxy/src/api_server_tcp.cc
tszumski: done

media-proxy/src/session-base.cc
tszumski: fixed

media-proxy/src/session-mtl-st2110-20.cc
tszumski: 1. Throw removed, 
2. mtl/rdma/memif initialization moved to init() method

tszumski:
> > I'm thinking what to do with error handling.
> > 
> > * In some places there is simply `return -1` used. I don't like it.
> > * Elsewhere, some semi-useful error codes ( like `return -EINVAL` ) are returned. That's better, the best what can be done in C, maybe only defining more descriptive err codes could improve the situation in C
> > * But this is c++ and sometimes throw is used, which seems to be the most elegant way. maybe slightly less performant, but not much, I guess it is the best everywhere besides critical paths.
> > 
> > Any thoughts? Are there any drawback of using try-catch and throw everywhere?
> 
> Consider returning errors, not throwing exceptions.

Done

media-proxy/src/api_server_tcp.cc
Sakoram: doesn't look like done

media-proxy/src/session-rdma-rx.cc
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    if (frame_thread_handle) {
+        frame_thread_handle->join();
+        delete frame_thread_handle;
+        frame_thread_handle = 0;
+    }
+}
+
+int RxRdmaSession::on_connect_cb(memif_conn_handle_t conn)
```
Sakoram, line 148: Please add empty method on_disconnect_cb in RxRdma and Tx RDMA and write inside the comment:
/* TODO: unregister in libfabric memory regions allocated by memif */
Thanks.

Now it is not an issue as we disconnect only when session is destroyed, but in case when we would like to disconnect and connect different memif connection to existing session it will be a problem.

media-proxy/src/session-rdma-rx.cc
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return 0;
+}
+
+RxRdmaSession::~RxRdmaSession()
+{
+    INFO("%s, fb_recv %d\n", __func__, fb_recv);
+    stop = true;
+    if (ep_ctx) {
+        ep_destroy(&ep_ctx);
```
Sakoram, line 132: This function can return an error, for now not very meaningful error but still, I think you should check it.

media-proxy/src/session-rdma-tx.cc
```diff
@@ -0,0 +1,137 @@
...
+    memif_region_details_t region;
+
+    int err = memif_get_buffs_region(conn, &region);
+    if (err) {
+        ERROR("%s, Getting memory buffers from memif failed. \n", __func__);
+        return err;
+    }
+
+    err = ep_reg_mr(ep_ctx, region.addr, region.size);
+    if (errno) {
```
Sakoram, line 131:
```suggestion
    if (err) {
```

media-proxy/src/session-rdma-rx.cc
```diff
@@ -0,0 +1,165 @@
...
+    memif_region_details_t region;
+
+    int err = memif_get_buffs_region(conn, &region);
+    if (err) {
+        ERROR("%s, Getting memory buffers from memif failed. \n", __func__);
+        return err;
+    }
+
+    err = ep_reg_mr(ep_ctx, region.addr, region.size);
+    if (errno) {
```
Sakoram, line 159:
```suggestion
    if (err) {
```

Sakoram (changes requested):
Resources already allocated by init() methods are not deallocated if the method fails on a next allocation. you just return -1 and create memory leak.

media-proxy/src/session-rdma-rx.cc
Sakoram: the same in tx :)

media-proxy/src/session-base.cc
Sakoram: I suggest to do it with std::jthread, std::stop_source and std::stop_token

ko80: https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/245

tszumski: memif_stop added in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/e1a5864a0a6f12017139869a11f5089261321e18

media-proxy/src/session-mtl-st2110-20.cc
tszumski: cleanup() no longer exist, destructor se stop variable and join threads

media-proxy/src/session-rdma-rx.cc
tszumski: fix in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/2969a18f4536ff15a4531a07715199489671437e

media-proxy/src/session-rdma-tx.cc
tszumski: fix in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/2969a18f4536ff15a4531a07715199489671437e

media-proxy/src/session-rdma-rx.cc
tszumski: fix in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/3d694d9aae84c488290fdbf5558135eea4dc2b21

media-proxy/src/session-rdma-rx.cc
tszumski: fix in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/3d694d9aae84c488290fdbf5558135eea4dc2b21

media-proxy/src/api_server_tcp.cc
tszumski: fix in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/36026af73506c0b23e4928f80afa3aaebd361e8f

tszumski:
> Resources already allocated by init() methods are not deallocated if the method fails on a next allocation. you just return -1 and create memory leak.

Good catch, fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/240/commits/1a4aa0997de39e956658e55828d9015ce27663e2

Sakoram (approved):
Still a few things to do, mainly we spotted threads synchronization issues, but...

This batch of improvements looks great,
Excellent job Tomek

PR #241, opened 2024-10-29
Fix RDMA support in sample apps

ko80:
This pull request includes the following commits:
* Fix RDMA support in sample apps
* Correct out_dir in text-rdma.sh

### DIFF
--- a/sdk/samples/recver_app.c
+++ b/sdk/samples/recver_app.c
@@ -217,35 +217,52 @@ int main(int argc, char** argv)
             goto error_delete_conn;
         }
     } else {
-        MeshConfig_ST2110 cfg;
+        if (!strcmp(payload_type, "rdma")) {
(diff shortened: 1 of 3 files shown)


ko80:
> I see some imperfections in the design as we are forced to create a big if statement to switch between payload types and set exactly the same fields (ip and ports) in almost identical structures.
> 
> I guess it should be done differently for example with some inheritance ( translated into C: placing common structure at the front of the specific structure) then we could set common parameters ones, and branch only to do things that are different.
> 
> Anyway, the code works. The design improvements can be done in the future.

Yeah, it will be reworked shortly since we've already planned to move to the JSON-based configuration.

Mionsz (approved):

tszumski (approved):
LGTM

awilczyns (approved):

PR #245, opened 2024-10-31
New Media Proxy architecture to support Multipoint Groups

ko80:
* Add documentation on Multipoint Groups architecture
* Enable C++20, add concurrency support and connection base class

### DIFF
--- a/media-proxy/include/mesh/CONCEPT.md
+++ b/media-proxy/include/mesh/CONCEPT.md
@@ -1,0 +1,269 @@
+# Concept – New Architecture of Media Proxy to Support Multipoint Groups
+
+## Connection (base class)
+* Interface for Transmitting data – _may be implemented in the derived class_.
+* Interface for Receiving data – _may be implemented in the derived class_.
+* State – _When changes, triggers an Event_.
+  * `Not Configured` – _After initialization._
+  * `Configured`
+  * `Establishing`
+  * `Active`
+  * `Suspended`
+  * `Closing`
+  * `Closed`
+  * `Deleting`
+* Operations
+  * `Create` – _Performed by constructor_.
+  * `Configure`
+  * `Establish`
+  * `Suspend` – _For test/debug purposes_.
+  * `Resume` – _For test/debug purposes_.
+  * `Shutdown`
+  * `Delete` – _Initiated externally. Can't be performed by the Connection class_.
+* Status – _When changes, triggers an Event_.
+  * `Initial`
+  * `Transition`
+  * `Healthy`
+  * `Failure`
+  * `Shutdown`
+* Base metrics
+  * Throughput – _Move calculation to the Telemetry extractor level_.
+  * Transaction counter.
+  * Byte counter.
+  * Error counter.
+  * Status.
+  * State.
+  * Maciej suggested:
+    * Backend load / capacity metric. Report example: `load: 85%` – _Secondary_.
+    * Interface to enable/disable metrics by name.
+  * Tomasz suggested:
+    * Unified logger interface.
+* Information
+  * Created at.
+  * Established at.
+* Interface for sending metrics.
+* Metrics Reset interface – _Access from MCM Agent, option to clear all or certain metrics_.
+
+### Intermediary Layer: ST2110, RDMA, Local
+* Base functionality of the specific backend.
+
+### Backend Layer: (xxx)-Tx, (xxx)-Rx
+
+* Transmitter or Receiver functionality of the specific backend.
+
+## Connection Lifecycle State Diagram
+
+```mermaid
+stateDiagram
+    notinit: Not Initialized
+    configuring: Configuring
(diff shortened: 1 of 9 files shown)


media-proxy/include/mesh/concurrency.h
```diff
@@ -0,0 +1,215 @@
...
+ * Channel
+ * 
+ * A thread-safe queue template supporting cancellation of blocking calls
+ * put() and get() by checking context::Context. To set a timeout on the
+ * blocking calls, context::WithTimeout is supposed to be used.
+ * Example:
+ * ...
+ */
+template<typename T>
+class Channel {
```
Sakoram, line 149: what is a use case for the channel in our project? Do you want to pass pointer to buffers between connections?

media-proxy/include/mesh/DEVELOPMENT.md
```diff
@@ -0,0 +1,195 @@
...
+    auto emulated_rx = new EmulatedReceiver;
+    emulated_rx->establish(ctx);
+
+    // Setup Rx connection
+    auto conn_rx = new mesh::connection::ST2110_20Rx;
+
+    res = conn_rx->configure(ctx, some_ST2110_20Rx_related_config);
+    if (res) {
+        printf("Configure Rx failed: %s\n", mesh::connection::result2str(res));
+        goto exit;
```
Sakoram, line 91: here you go from `Not Initialized` to shutdown, but your Lifecycle State Diagram does not show this flow.

I guess we need to be able to go from every state to `Deleting`

media-proxy/include/mesh/concurrency.h
ko80: Channels are thread safe queues. They will be used to deliver events within Media Proxy:
* Connection created
* Connection destroyed
* Bridge created
* Bridge destroyed
* Multipoint Group created
* Multipoint Group destroyed
* Other events related to transitions between internal states.
Another uses case for Channel is a wait group (see `sync.WaitGroup` in Golang). It allows to control the flow and wait until some processing in parallel threads is completed. I'll show it later in code samples.

media-proxy/include/mesh/DEVELOPMENT.md
ko80: Agree

media-proxy/include/mesh/concurrency.h
Sakoram: What if we implement it as a lock free queue? (for example with [this](https://www.boost.org/doc/libs/1_86_0/doc/html/lockfree.html))  To be honest, I'm not sure how this will affect performance and whether we need good performance in events at all, but I just want to throw my thought out there.

ko80: Events will never be generated in the hot path of Data Plane. We don't need tangible performance in Control Plane at all. The idea of a lock-free queue is good, however it will be way more complex because we can't combine the `condition_variable` waiting with lock-free primitives like atomics. We can't use the C++20 `wait` of atomics because we can't cancel it. I even can't measure the complication that the lock-free implementation could bring here.

So, let's don't optimize prematurely ;-)

Sakoram: agree, lack of `wait` would be an issue.

tszumski (approved):

media-proxy/include/mesh/CONCEPT.md
```diff
@@ -0,0 +1,271 @@
...
+* Transmitter or Receiver functionality of the specific backend.
+
+## Connection Lifecycle State Diagram
+
+```mermaid
+stateDiagram
+    notinit: Not Initialized
+    configuring: Configuring
+    configured: Configured
+    establishing: Establishing
```
Sakoram, line 61: do we really need those transition states like: Configuring, Establishing, Deleting? Can't we go directly to configured, established etc. ?

ko80: Let's take a look at the `Configure` flow. The base class initiates a transition by setting the state to `Configuring`. Then, the derived class is responsible for the actual action to apply a configuration. According to the result of that action, the derived class sets the state to `Configured` or `Not Initialized`.

Now the `Establish` flow. The base class initiates a transition by setting the state to `Establishing`. Then, the derived class is responsible for the actual action to establish a connection. According to the result of that action, the derived class sets the state to `Active` or `Closed`.

The mentioned actions in the derived class can consume time, few moments or seconds. During that period, the current state will represent the fact that the action is being taken. So, yes, we need those transition states.

Sakoram: fair enough

Sakoram (approved):
I am not 100% sure that reusing patterns straight from Golang in c++ is the best approach.

I'm guessing that some c++ patterns exist that fit our case, but I'm not able to propose something better, so we can try to go with this proposition.

PR #248, opened 2024-11-07
Add pingpong apps capable of measuring latency

Sakoram:
The code quality of pingpong apps is NOT production grade, but they work

- added a mutex to protect vector containing session contexts in proxy_context.cc.

### DIFF
--- a/media-proxy/src/proxy_context.cc
+++ b/media-proxy/src/proxy_context.cc
@@ -186,6 +186,7 @@ int ProxyContext::RxStart_rdma(const mcm_conn_param *request)
     }
 
     INFO("%s, session id: %d", __func__, session_ptr->get_id());
+    std::lock_guard<std::mutex> lock(ctx_mtx);
     mDpCtx.push_back(session_ptr);
     return session_ptr->get_id();
 }
@@ -251,6 +252,7 @@ int ProxyContext::RxStart_mtl(const mcm_conn_param *request)
     }
 
     INFO("%s, session id: %d", __func__, session_ptr->get_id());
+    std::lock_guard<std::mutex> lock(ctx_mtx);
     mDpCtx.push_back(session_ptr);
     return session_ptr->get_id();
 }
@@ -287,6 +289,7 @@ int ProxyContext::TxStart_rdma(const mcm_conn_param *request)
     }
 
     INFO("%s, session id: %d", __func__, session_ptr->get_id());
+    std::lock_guard<std::mutex> lock(ctx_mtx);
     mDpCtx.push_back(session_ptr);
     return session_ptr->get_id();
 }
@@ -351,6 +354,7 @@ int ProxyContext::TxStart_mtl(const mcm_conn_param *request)
     }
 
     INFO("%s, session id: %d", __func__, session_ptr->get_id());
+    std::lock_guard<std::mutex> lock(ctx_mtx);
     mDpCtx.push_back(session_ptr);
     return session_ptr->get_id();
 }
@@ -366,6 +370,7 @@ int ProxyContext::TxStart(const mcm_conn_param *request)
 int ProxyContext::Stop(const int32_t session_id)
 {
     int ret = 0;
+    std::lock_guard<std::mutex> lock(ctx_mtx);
     auto ctx = std::find_if(mDpCtx.begin(), mDpCtx.end(),
                             [session_id](auto it) { return it->get_id() == session_id; });
     if (ctx != mDpCtx.end()) {
--- a/sdk/samples/ping_app.c
+++ b/sdk/samples/ping_app.c
@@ -1,0 +1,399 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024 Intel Corporation
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#define _GNU_SOURCE
+#include <assert.h>
+#include <bsd/string.h>
+#include <getopt.h>
+#include <linux/limits.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <time.h>
+#include <unistd.h>
+#include "mesh_dp.h"
+#include <pthread.h>
+#include <stdatomic.h>
+#include <sched.h>
+#include "pingpong_common.h"
+
+#define DEFAULT_RECV_PORT "9001"
+#define DEFAULT_SEND_PORT "10001"
+
+static volatile bool keepRunning = true;
+
+static MeshClient *client;
+atomic_int counter = 0;
+
+
+Config config = {
+    .recv_addr = DEFAULT_RECV_IP,
+    .recv_port = DEFAULT_RECV_PORT,
+    .send_addr = DEFAULT_SEND_IP,
+    .send_port = DEFAULT_SEND_PORT,
+    
+    .payload_type = "",
+    .protocol_type = "",
+    .pix_fmt_string = DEFAULT_VIDEO_FMT,
+    .socket_path = DEFAULT_MEMIF_SOCKET_PATH,
+    .interface_id = DEFAULT_MEMIF_INTERFACE_ID,
+    .loop = DEFAULT_INFINITE_LOOP,
+
+    .width = DEFAULT_FRAME_WIDTH,
+    .height = DEFAULT_FRAME_HEIGHT,
+    .vid_fps = DEFAULT_FPS,
+    .frame_size = 0,
+    .total_num = 300
+};
+
+void intHandler(int dummy)
+{
+    keepRunning = 0;
+}
+
+/* print a description of all supported options */
+void usage(FILE *fp, const char *path)
+{
+    /* take only the last portion of the path */
+    const char *basename = strrchr(path, '/');
+    basename = basename ? basename + 1 : path;
+
+    fprintf(fp, "usage: %s [OPTION]\n", basename);
+    fprintf(fp, "-H, --help\t\t\t"
+                "Print this help and exit\n");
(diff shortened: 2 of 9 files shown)


sdk/samples/pong_app.c
```diff
@@ -0,0 +1,331 @@
...
+#include "mesh_dp.h"
+#include <pthread.h>
+#include <stdatomic.h>
+#include <sched.h>
+#include "pingpong_common.h"
+
+#define DEFAULT_RECV_IP "127.0.0.1"
+#define DEFAULT_RECV_PORT "10001"
+#define DEFAULT_SEND_IP "127.0.0.1"
+#define DEFAULT_SEND_PORT "9001"
```
tszumski, line 28: Move those define to *common.h and unify them between ping/pong

sdk/samples/pong_app.c
```diff
@@ -0,0 +1,331 @@
...
+    
+    // Set CPU affinity
+    CPU_ZERO(&cpuset);
+    CPU_SET(thread_id + config.threads_num * 2 % CPU_CORES, &cpuset);  // Pin to specific CPU core
+    if (pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset) != 0) {
+        perror("pthread_setaffinity_np");
+        pthread_exit(NULL);
+    }
+
+    for (int i = 0; true; i++) {
```
tszumski, line 153: while(true) instead of for-loop ?

media-proxy/src/proxy_context.cc
```diff
@@ -664,6 +665,7 @@ int ProxyContext::RxStart_mtl(const mcm_conn_param *request)
     std::cout << "session id: " << st_ctx->id << std::endl;
     INFO("%s, session id: %d", __func__, st_ctx->id);
     st_ctx->type = RX;
+    std::lock_guard<std::mutex> lock(ctx_mtx);
```
tszumski, line 668: mDpCtx is also used in api_server_tcp.cc line 133, 215 and 244
I think lock should be applied there too

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,400 @@
...
+
+    // Set CPU affinity
+    CPU_ZERO(&cpuset);
+    CPU_SET(thread_data->thread_id + config.threads_num % CPU_CORES, &cpuset);  // Pin to specific CPU core
+    if (pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset) != 0) {
+        perror("pthread_setaffinity_np");
+        pthread_exit(NULL);
+    }
+
+    for (int i = 0; true; i++) {
```
tszumski, line 148: Same here, please change to while(true)

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,400 @@
...
+        mesh_delete_connection(&conn);
+        pthread_exit(NULL);
+    }
+
+    if (thread_data->thread_id == 0)
+        config.frame_size = conn->buf_size;
+
+    // Set CPU affinity
+    CPU_ZERO(&cpuset);
+    CPU_SET(thread_data->thread_id + config.threads_num % CPU_CORES, &cpuset);  // Pin to specific CPU core
```
tszumski, line 142: Please add brackets: (thread_data->thread_id + config.threads_num) % CPU_CORES

What is the reason to pin threads to specific cores? You should consider pinning all threads to single socket (that is directly connected to NIC you want to use)

sdk/samples/pong_app.c
```diff
@@ -0,0 +1,331 @@
...
+    if (err) {
+        printf("ERROR: init_conn failed \n");
+        mesh_delete_connection(&s_conn);
+        pthread_exit(NULL);
+    }
+
+    
+    // Set CPU affinity
+    CPU_ZERO(&cpuset);
+    CPU_SET(thread_id + config.threads_num * 2 % CPU_CORES, &cpuset);  // Pin to specific CPU core
```
tszumski, line 147: Add brackets:
(thread_id + config.threads_num * 2) % CPU_CORES

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,400 @@
...
+    double *latency_results;
+}t_data;
+
+void *receiver_thread(void *arg) {
+    t_data *thread_data = (t_data *)arg;
+    struct timespec send_time = {}, now = {};
+    int recved_pkt_num;
+    cpu_set_t cpuset;
+    int err;
+    int timeout = -1;
```
tszumski, line 118: Please rename timeout to timeout_ms for code readability

sdk/samples/pong_app.c
```diff
@@ -0,0 +1,331 @@
...
+                "Set infinite loop sending (default: %d)\n",
+        DEFAULT_INFINITE_LOOP);
+    fprintf(fp, "\n");
+}
+
+void *pong_thread(void *arg) {
+    int thread_id = *(int *)arg;
+    cpu_set_t cpuset;
+    int err;
+    int timeout = -1;
```
tszumski, line 108: rename timeout to timeout_ms

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,400 @@
...
+        usleep(100000);
+    }
+
+    // time to setup
+    sleep(1);
+
+    // Increment atomic counter and signal threads
+    for (int i = 1; i <= PRINT_COUNT; i++) {
+        usleep(500000);  // Delay to synchronize output
+        atomic_store(&counter, i);
```
tszumski, line 375: Consider "counter" to be unique for each thread, to make connection more "natural" right now you will see network bw spikes

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,400 @@
...
+        }
+        usleep(100000);
+    }
+
+    // time to setup
+    sleep(1);
+
+    // Increment atomic counter and signal threads
+    for (int i = 1; i <= PRINT_COUNT; i++) {
+        usleep(500000);  // Delay to synchronize output
```
tszumski, line 374: Delay should be based on vid_fps

sdk/samples/ping_app.c
Sakoram: This is the goal. We want to test latency in the function of number of concurrent session sending buffer at the same time

sdk/samples/ping_app.c
Sakoram: ok, it can be, but here this parameter will serve different purpose. It those tests I just wanted to test latency without using full bandwidth, I just put big magic number as a delay between buffers, but I can parametrize this and allow the user to specify this

sdk/samples/pong_app.c
Sakoram: I moved IPs, but ports have to be different in each app so I left them

sdk/samples/pong_app.c
Sakoram: changed.

media-proxy/src/proxy_context.cc
Sakoram: done

sdk/samples/ping_app.c
Sakoram: done

sdk/samples/ping_app.c
Sakoram: brackets added

sdk/samples/pong_app.c
Sakoram: added

sdk/samples/ping_app.c
Sakoram: done

sdk/samples/ping_app.c
Sakoram: done

sdk/samples/ping_app.c
Sakoram: numactl --cpunodebind=0 --membind=0 seems to break the connection. idk why, but I won't use it

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,399 @@
...
+        1);
+    fprintf(fp, "\n");
+}
+
+typedef struct {
+    int thread_id;
+    double *latency_results;
+} t_data;
+
+void *receiver_thread(void *arg)
```
ko80, line 110:
```suggestion
void * receiver_thread(void *arg)
```

sdk/samples/ping_app.c
```diff
@@ -0,0 +1,399 @@
...
+    int err;
+    int timeout_ms = -1;
+
+    MeshConnection *conn;
+    MeshBuffer *buf;
+
+    err = mesh_create_connection(client, &conn);
+    if (err) {
+        printf("Failed to create a mesh connection: %s (%d)\n", mesh_err2str(err), err);
+        pthread_exit(NULL);
```
ko80, line 125: Why can't you simply do `return;` here? Do you return something except `NULL` from these thread functions?

Sakoram: We can, I think both pthread_exit() and return from thread function does the same thing, pthread_exit seemed more explicit so I used this, but I can change it

ko80: Yeah, please replace with `return` where applicable.

sdk/samples/ping_app.c
Sakoram: done

sdk/samples/ping_app.c
Sakoram: I did remove it, but just a comment, that pthread_exit is also used in `api_server_tcp.cc`

ko80: `api_server_tcp.cc` is to be excluded from the project quite soon. Don't refer to it's code.

tszumski (approved):

ko80 (approved):

PR #253, opened 2024-11-13
Fix multiple Coverity issues

tszumski:
1. Add initialization of source_begin_iova_map_sz, shm_buf_num
2. Pass dev_port by reference 
3.  Apply lock before accesing fifo.empty()

### DIFF
--- a/media-proxy/include/session-mtl.h
+++ b/media-proxy/include/session-mtl.h
@@ -60,7 +60,7 @@ class RxSt20MtlSession : public MtlSession
     size_t source_begin_iova_map_sz;
 #endif
 
-    void copy_connection_params(const mcm_conn_param &request, std::string dev_port);
+    void copy_connection_params(const mcm_conn_param &request, std::string &dev_port);
     void consume_frame(struct st_frame *frame);
     void frame_thread();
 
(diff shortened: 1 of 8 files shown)


media-proxy/include/session-mtl.h
```diff
@@ -60,7 +60,7 @@ class RxSt20MtlSession : public MtlSession
     size_t source_begin_iova_map_sz;
 #endif
 
-    void copy_connection_params(const mcm_conn_param &request, std::string dev_port);
+    void copy_connection_params(const mcm_conn_param &request, std::string &dev_port);
```
ko80, line 63: C++ reference style: the ampersand sticks to the argument type definition. 
```suggestion
    void copy_connection_params(const mcm_conn_param& request, std::string& dev_port);
```

Consider updating that in all your work.

tszumski: Then .clang-format is not configured correctly.
Can we postpone it for later? Fix .clang-format and apply format globally?

ko80: Of course, I have no objections.

ko80 (dismissed):

Sakoram (approved):
Looks good,

only comment: consider using std::lock_guard instead of standard mutex

PR #257, opened 2024-11-15
Fix development.md

tszumski:
Fix development.md : 
1. Set state to configured in EmulatedReceiver and EmulatedTransmitter constructor
2.  Add transmit_wrapper() for EmulatedTransmitter

### DIFF
--- a/media-proxy/include/mesh/DEVELOPMENT.md
+++ b/media-proxy/include/mesh/DEVELOPMENT.md
@@ -59,7 +59,11 @@ using namespace mesh;
 
 class EmulatedReceiver : public connection::Connection {
 public:
-    EmulatedReceiver() : _kind(connection::Kind::receiver) {}
+    EmulatedReceiver(context::Context &ctx)
+    {
+        _kind = connection::Kind::receiver;
+        set_state(ctx, connection::State::configured);
+    }
 
     connection::Result on_establish(context::Context& ctx) {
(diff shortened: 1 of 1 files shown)


media-proxy/include/mesh/conn.h
```diff
@@ -92,6 +92,8 @@ class Connection {
     Result do_receive(context::Context& ctx, void *ptr, uint32_t sz,
                       uint32_t& sent);
 
+    Result transmit(context::Context& ctx, void *ptr, uint32_t sz);
```
ko80, line 95: Don't make it public. This method has intentionally been made protected to allow calling it only from classes derived from Connection.

Add an additional public wrapper method to the emulated tx/rx to call `trasmit()`.

ko80 (changes requested):

media-proxy/include/mesh/DEVELOPMENT.md
```diff
@@ -71,6 +71,11 @@ public:
                                   uint32_t& sent) {
         printf("Received %s (%u)\n", (char *)ptr, sz);
     }
+
```
ko80, line 74: Don't add a new method for that. Set the `configured` state in the emulated tx/rx constructor instead.

ko80 (changes requested):

ko80 (approved):

PR #258, opened 2024-11-17
Add Unit Tests for libfabric-based RDMA implementation

Sakoram:
- Use `struct libfabric_*_ops_t` structures with function pointers to access every function in libfabric-based RDMA implementation. This allows good mocking. Sadly, I haven't found any better solution to link unit tests with media-proxy and be able to test in one test file and mock in another test file the same functions from media-proxy. I'm open to suggestions how to do it better.
 - Add unit tests for libfabric-based RDMA implementation
 - Fix small bugs in libfabric-based RDMA implementation
 - Remove rdma hmem (It was added to potentially support RDMA transfers directly to GPU memory. However, it was never used and probably higher level code should be used to enable this if we want to achieve it in the future)


moleksy:
At first glance there won't be a lot of code rework on my side, I should be able to do it in an hour or so with some testing.

ko80:
This is to be merged after https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/259
@Sakoram please rebase.

ko80 (approved):

PR #259, opened 2024-11-18
Add base memif local connection support for Multipoint Groups

ko80:
* Add gRPC API between SDK and Media Proxy.
* Move the Protobuf .proto files to a unique top level directory.
* Add Memif connection classes to support Multipoint Groups.
* Add a manager to handle memif connections.
* Refactor SDK API implementation to allow using gRPC instead of TCP if a configuration flag is set.
* Add structured logger supporting standard and JSON output.
* Split unit tests to isolate SDK from Media Proxy.
* Refactor the build system to accommodate the changes above.
* Update Docker build.


build.sh
```diff
@@ -28,4 +28,8 @@ run_as_root_user ln -s /usr/lib64/libbpf.so.1 /usr/lib/x86_64-linux-gnu/libbpf.s
 run_as_root_user ldconfig
 
 # Run unit tests
-LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib" "${BUILD_DIR}/bin/mcm_unit_tests"
+export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib"
+"${BUILD_DIR}/bin/sdk_unit_tests"
+"${BUILD_DIR}/bin/media_proxy_unit_tests"
+
+echo -e "\e[48;2;0;255;0;30m Build Succeeded \e[0m"
```
Mionsz, line 35: Could you change the build.sh in next PR and just use one of methods that are implemented and included in the scriipt:
 
image
 
Instead of this .
 
you can just write: prompt Build Succeeded

image
 
it will be blue but I can add the succeeded method with proper colors - the above can have be disabled here: https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/main/scripts/common.sh
 
so that it can be done globally for placess where it should

Mionsz (dismissed):

build.sh
ko80: Good idea! Will do in the next PR.

moleksy (approved):

Sakoram (approved):

media-proxy/src/mesh/conn_local.cc
```diff
@@ -0,0 +1,220 @@
...
+
+    bzero(memif_socket_args.app_name, sizeof(memif_socket_args.app_name));
+    bzero(memif_socket_args.path, sizeof(memif_socket_args.path));
+
+    if (!ops) {
+        ops = &this->ops;
+        default_memif_ops(ops);
+    }
+
+    // Set application name
```
Sakoram, line 27: please use `/* */` type of comments in this file.

Sakoram:
I don't know how ready is this code, but you often use `//` type of comments instead of `/* */`. 
Also you left a few commented lines of code.

media-proxy/src/api_server_tcp.cc
```diff
@@ -251,8 +252,8 @@ void RunTCPServer(ProxyContext* ctx)
         return;
     }
 
-    INFO("TCP Server listening on %s", ctx->getTCPListenAddress().c_str());
-    registerSignals();
+    mesh::log::info("TCP Server listening on %s", ctx->getTCPListenAddress().c_str());
+    // registerSignals();
```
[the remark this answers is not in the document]
ko80: The signals are handled in `media_proxy.cc` line 131, 132,

ko80:
> I don't know how ready is this code, but you often use `//` type of comments instead of `/* */`.
C++ style requires `//` instead of `/* */`. I decided to stick to `//` mostly but in some places still use `/** ..... */`.

> Also you left a few commented lines of code.
That's possible. Let me remove that in the next PRs.

tszumski (dismissed):

Sakoram:
> > I don't know how ready is this code, but you often use `//` type of comments instead of `/* */`.
> 
> C++ style requires `//` instead of `/* */`. I decided to stick to `//` mostly but in some places still use `/** ..... */`.
> 
> > Also you left a few commented lines of code.
> 
> That's possible. Let me remove that in the next PRs.

I think we should just stick to one style in all c and c++ files. We will not use such archaic compilers , so that they do not support ‘//’ style comments in C, so we can for example use `/* */` for multiline comments and `//` for single line comments

media-proxy/src/api_server_tcp.cc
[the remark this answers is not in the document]
Sakoram: alright, I missed that. Please remove code responsible for handling signals in this file.

media-proxy/src/mesh/conn.cc
```diff
@@ -208,6 +211,11 @@ Result Connection::set_link(context::Context& ctx, Connection *new_link)
     return set_result(Result::success); // TODO: return error if needed
 }
 
+Connection * Connection::link()
```
Sakoram, line 214:
```suggestion
Connection * Connection::get_link()
```
I think getter is more readable when it has `get_` prefix

media-proxy/src/mesh/uuid.cc
```diff
@@ -0,0 +1,47 @@
...
+#include <iomanip>
+#include <mutex>
+
+namespace mesh {
+
+std::string generate_uuid_v4()
+{
+    static std::random_device rd;
+    static std::mt19937_64 gen(rd());
+    static std::uniform_int_distribution<uint64_t> dis;
```
Sakoram, line 18:
```suggestion
    static std::uniform_int_distribution<uint8_t> dis;
```

media-proxy/src/api_server_tcp.cc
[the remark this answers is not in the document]
ko80: I'm gonna keep that commented out for a while because the work on graceful shutdown is not completed yet. Will remove that when done.

ko80:
> I think we should just stick to one style in all c and c++ files. We will not use such archaic compilers , so that they do not support ‘//’ style comments in C, so we can for example use `/* */` for multiline comments and `//` for single line comments

Since we don't have requirements to the SDK API, we must support the earlier possible version of C in it. I believe we must follow the strongest rules used in Linux kernel and drivers. But this affects only the frontend part `mesh_dp.h` for sure. I don't mind using `/*  */` in C++ at all.

For now, let's keep this as it is. We can refactor that in the future if you'd like.

media-proxy/src/mesh/conn.cc
ko80: Following your suggestion, we'll have to change all similar method names:
```cpp
Kind kind();
State state();
Status status();
Connection * link();
```
this way:
```cpp
Kind get_kind();
State get_state();
Status get_status();
Connection * get_link();
```
Personally, I don't like it. It seems like a C# or Java approach to name getters/setters. I like the `state() / set_state()` way of giving names much more because it's less words, and it's the Go style :-)

tszumski (approved):

media-proxy/include/mesh/concurrency.h
```diff
@@ -36,19 +36,31 @@ namespace context {
...
     std::stop_token stop_token();
     bool done();
 
-protected:
     std::stop_source ss;
-    thread::Channel<bool> *ch;
 
+protected:
     Context();
+    Context(Context& parent);
```
tszumski, line 50: Following out conversation, please move default constructor to public

tszumski (approved):
LGTM

Sakoram (dismissed):

media-proxy/src/mesh/concurrency.cc
```diff
@@ -11,19 +11,87 @@ namespace mesh {
...
+                     parent(nullptr),
+                     cb(nullptr),
+                     ch(new thread::Channel<bool>(1)),
+                     timeout_ms(std::chrono::milliseconds(0))
 {
 }
 
+Context::Context(Context& parent) : ss(std::stop_source()),
+                                    parent(&parent),
+                                    ch(new thread::Channel<bool>(1))
```
tszumski, line 24: For `Context::Context(Context& parent)` please also initialize `timeout_ms`
`timeout_ms(std::chrono::milliseconds(0))`

Otherwise there is usage fn uninitialized value

tszumski (changes requested):

media-proxy/src/mesh/concurrency.cc
ko80: Fixed https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/259/commits/e1e36e2f679761baefd657ffb9310d7959cf5637

build.sh
ko80: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/1f11bbc0e781d91348d91e782c09d1d35e2c87ee

tszumski (approved):

Mionsz (approved):

PR #261, opened 2024-11-21
Add ST2110 bridge classes derived from Connection

tszumski:
- Add new error code "error_shutdown",
- Fix use of uninitialize value _link,
- Fix clang compilation error - uuid.cc
- Add ST2110 Rx/Tx 20/22/30 connections with basic UT

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,7 +3,11 @@ BasedOnStyle: LLVM
 IndentWidth: 4
 UseTab: Never
 ColumnLimit: 100
-BreakBeforeBraces: Linux
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: false
+  AfterNamespace: false
+  AfterFunction: true
 AllowShortIfStatementsOnASingleLine: false
 AllowShortEnumsOnASingleLine: false
 IndentCaseLabels: false
--- a/.clang-format
+++ b/.clang-format
@@ -7,10 +7,11 @@ BreakBeforeBraces: Custom
 BraceWrapping:
   AfterClass: false
   AfterNamespace: false
-  AfterFunction: true
+  AfterFunction: false
 AllowShortIfStatementsOnASingleLine: false
 AllowShortEnumsOnASingleLine: false
 IndentCaseLabels: false
 IncludeBlocks: Preserve
 SortIncludes: false
+ReferenceAlignment: Left
 ...
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -66,7 +66,7 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
-
+    error_shutdown,
     // TODO: more error codes to be added...
 };
 
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -66,7 +66,7 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
-    error_shutdown,
+    error_context_cancelled,
     // TODO: more error codes to be added...
 };
 
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -40,13 +40,12 @@ class ST2110 : public Connection {
                                      const char local_ip_addr[MESH_IP_ADDRESS_SIZE],
                                      int& session_id);
 
-    ST2110() : mtl_device(nullptr) {};
     virtual ~ST2110() {};
 
   protected:
     static int frame_available_cb(void *ptr);
 
-    mtl_handle mtl_device;
+    mtl_handle mtl_device = nullptr;
     std::atomic<bool> stop;
 };
 
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -20,7 +20,7 @@ namespace mesh::connection {
 /**
  * ST2110
  *
- * Base abstract class of ST2110. ST2110Rx/ST2110Tx
+ * Base abstract class of SPMTE ST2110-xx bridge. ST2110Rx/ST2110Tx
  * inherit this class.
  */
 class ST2110 : public Connection {
@@ -38,13 +38,13 @@ class ST2110 : public Connection {
     static mtl_handle get_mtl_handle(const std::string &dev_port, mtl_log_level log_level,
                                      const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
 
-    ST2110() : _st(0){};
+    ST2110() : mtl_device(nullptr) {};
     virtual ~ST2110(){};
 
   protected:
     static int frame_available_cb(void *ptr);
 
-    mtl_handle _st;
+    mtl_handle mtl_device;
     std::atomic<bool> _stop;
     std::condition_variable_any _cv;
     std::mutex _mx;
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -25,10 +25,11 @@ namespace mesh::connection {
  */
 class ST2110 : public Connection {
   public:
-    static st_frame_fmt mesh_video_format_to_st_format(int fmt);
-    static st30_fmt mesh_audio_format_to_st_format(int fmt);
-    static st30_sampling mesh_audio_sampling_to_st_sampling(int sampling);
-    static st30_ptime mesh_audio_ptime_to_st_ptime(int ptime);
+    static int mesh_video_format_to_st_format(int fmt, st_frame_fmt &st_fmt);
+    static int mesh_audio_format_to_st_format(int fmt, st30_fmt &st_fmt);
+    static int mesh_audio_sampling_to_st_sampling(int sampling, st30_sampling &st_sampling);
+    static int mesh_audio_ptime_to_st_ptime(int ptime, st30_ptime &st_ptime);
+
     static void *get_frame_data_ptr(st_frame *src);
     static void *get_frame_data_ptr(st30_frame *src);
 
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -36,8 +36,9 @@ class ST2110 : public Connection {
     static void get_mtl_dev_params(mtl_init_params &st_param, const std::string &dev_port,
                                    mtl_log_level log_level,
                                    const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
-    static mtl_handle get_mtl_handle(const std::string &dev_port, mtl_log_level log_level,
-                                     const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
+    static mtl_handle get_mtl_device(const std::string &dev_port, mtl_log_level log_level,
+                                     const char local_ip_addr[MESH_IP_ADDRESS_SIZE],
+                                     int &session_id);
 
     ST2110() : mtl_device(nullptr) {};
     virtual ~ST2110(){};
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -17,39 +17,124 @@ namespace mesh::connection {
 #define ST_APP_PAYLOAD_TYPE_ST20 (112)
 #define ST_APP_PAYLOAD_TYPE_ST22 (114)
 
+int mesh_video_format_to_st_format(int fmt, st_frame_fmt& st_fmt);
+int mesh_audio_format_to_st_format(int fmt, st30_fmt& st_fmt);
+int mesh_audio_sampling_to_st_sampling(int sampling, st30_sampling& st_sampling);
+int mesh_audio_ptime_to_st_ptime(int ptime, st30_ptime& st_ptime);
+
+void *get_frame_data_ptr(st_frame *src);
+void *get_frame_data_ptr(st30_frame *src);
+
+void get_mtl_dev_params(mtl_init_params& st_param, const std::string& dev_port,
+                        mtl_log_level log_level, const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
+mtl_handle get_mtl_device(const std::string& dev_port, mtl_log_level log_level,
+                          const char local_ip_addr[MESH_IP_ADDRESS_SIZE], int& session_id);
+
 /**
  * ST2110
  *
  * Base abstract class of SPMTE ST2110-xx bridge. ST2110Rx/ST2110Tx
  * inherit this class.
  */
-class ST2110 : public Connection {
+template <typename FRAME, typename HANDLE, typename OPS> class ST2110 : public Connection {
   public:
-    static int mesh_video_format_to_st_format(int fmt, st_frame_fmt& st_fmt);
-    static int mesh_audio_format_to_st_format(int fmt, st30_fmt& st_fmt);
-    static int mesh_audio_sampling_to_st_sampling(int sampling, st30_sampling& st_sampling);
-    static int mesh_audio_ptime_to_st_ptime(int ptime, st30_ptime& st_ptime);
-
-    static void *get_frame_data_ptr(st_frame *src);
-    static void *get_frame_data_ptr(st30_frame *src);
-
-    static void get_mtl_dev_params(mtl_init_params& st_param, const std::string& dev_port,
-                                   mtl_log_level log_level,
-                                   const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
-    static mtl_handle get_mtl_device(const std::string& dev_port, mtl_log_level log_level,
-                                     const char local_ip_addr[MESH_IP_ADDRESS_SIZE],
-                                     int& session_id);
-
-    virtual ~ST2110() {};
+    virtual ~ST2110() {
+        shutdown(_ctx);
+        if (ops.name)
+            free((void *)ops.name);
+    };
 
   protected:
-    static int frame_available_cb(void *ptr);
-    void init_frame_available();
-    void notify_frame_available();
-    void wait_frame_available();
-
     mtl_handle mtl_device = nullptr;
+    HANDLE mtl_session = nullptr;
+    OPS ops = {0};
+    size_t transfer_size = 0;
     std::atomic<bool> frame_available;
+    context::Context _ctx = context::WithCancel(context::Background());
+
+    virtual FRAME *get_frame(HANDLE) = 0;
+    virtual int put_frame(HANDLE, FRAME *) = 0;
+    virtual HANDLE create_session(mtl_handle, OPS *) = 0;
+    virtual int close_session(HANDLE) = 0;
+
+    static int frame_available_cb(void *ptr) {
+        auto _this = static_cast<ST2110 *>(ptr);
+        if (!_this) {
+            return -1;
+        }
+
+        _this->notify_frame_available();
+
+        return 0;
+    }
+
+    void notify_frame_available() {
+        frame_available.store(true, std::memory_order_release);
+        frame_available.notify_one();
+    }
+
+    void wait_frame_available() {
+        frame_available.wait(false, std::memory_order_acquire);
+        frame_available = false;
+    }
+
+    virtual int configure_common(context::Context& ctx, const std::string& dev_port,
+                                 const MeshConfig_ST2110& cfg_st2110) {
+        int session_id = 0;
+        mtl_device = get_mtl_device(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr, session_id);
+        if (!mtl_device) {
+            log::error("Failed to get MTL device");
+            return -1;
+        }
+
+        strlcpy(ops.port.port[MTL_PORT_P], dev_port.c_str(), MTL_PORT_MAX_LEN);
+        ops.port.num_port = 1;
+
+        char session_name[NAME_MAX] = "";
+        snprintf(session_name, NAME_MAX, "mcm_mtl_%d", session_id);
+        if (ops.name)
+            free((void *)ops.name);
+        ops.name = strdup(session_name);
+        ops.framebuff_cnt = 4;
+
+        ops.priv = this; // app handle register to lib
+        ops.notify_frame_available = frame_available_cb;
+
+        log::info("ST2110: configure")
+            ("port", ops.port.port[MTL_PORT_P])
+            ("num_port", (int)ops.port.num_port)
+            ("name", ops.name)
+            ("framebuff_cnt", ops.framebuff_cnt);
+
+        return 0;
+    }
+
+    Result on_establish(context::Context& ctx) override {
+        _ctx = context::WithCancel(ctx);
+        frame_available = false;
+
+        mtl_session = create_session(mtl_device, &ops);
+        if (!mtl_session) {
+            log::error("Failed to create session");
+            set_state(ctx, State::closed);
+            return set_result(Result::error_general_failure);
+        }
+
+        set_state(ctx, State::active);
+        return set_result(Result::success);
+    }
+
+    Result on_shutdown(context::Context& ctx) override {
+        _ctx.cancel();
+        notify_frame_available();
+
+        if (mtl_session) {
+            close_session(mtl_session);
+            mtl_session = nullptr;
+        }
+        set_state(ctx, State::closed);
+        return set_result(Result::success);
+    };
 };
 
 } // namespace mesh::connection
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -1,0 +1,63 @@
+#ifndef ST2110_H
+#define ST2110_H
+
+#include <thread>
+#include <bsd/string.h>
+#include <arpa/inet.h>
+#include <mtl/st_pipeline_api.h>
+#include <mtl/st30_pipeline_api.h>
+
+#include "conn.h"
+#include "mesh_dp.h"
+#include "logger.h"
+
+namespace mesh
+{
+
+namespace connection
+{
+
+#define ST_APP_PAYLOAD_TYPE_ST30 (111)
+#define ST_APP_PAYLOAD_TYPE_ST20 (112)
+#define ST_APP_PAYLOAD_TYPE_ST22 (114)
+
+/**
+ * ST2110
+ *
+ * Base abstract class of ST2110. ST2110Rx/ST2110Tx
+ * inherit this class.
+ */
+class ST2110 : public Connection
+{
+  public:
+    static st_frame_fmt mesh_video_format_to_st_format(int fmt);
+    static st30_fmt mesh_audio_format_to_st_format(int fmt);
+    static st30_sampling mesh_audio_sampling_to_st_sampling(int sampling);
+    static st30_ptime mesh_audio_ptime_to_st_ptime(int ptime);
+    static void *get_frame_data_ptr(st_frame *src);
+    static void *get_frame_data_ptr(st30_frame *src);
+
+    static void get_mtl_dev_params(mtl_init_params &st_param, const std::string &dev_port,
+                                   mtl_log_level log_level,
+                                   const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
+    static mtl_handle get_mtl_handle(const std::string &dev_port, mtl_log_level log_level,
+                                     const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
+
+    ST2110() : _st(0){};
+    virtual ~ST2110(){};
+
+  protected:
+    static int frame_available_cb(void *ptr);
+
+    mtl_handle _st;
+    std::atomic<bool> _stop;
+    std::condition_variable_any _cv;
+
+  private:
+};
+
+} // namespace connection
+
+} // namespace mesh
+
+#endif // ST2110_H
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -44,9 +44,12 @@ class ST2110 : public Connection {
 
   protected:
     static int frame_available_cb(void *ptr);
+    void init_frame_available();
+    void notify_frame_available();
+    void wait_frame_available();
 
     mtl_handle mtl_device = nullptr;
-    std::atomic<bool> stop;
+    std::atomic<bool> frame_available;
 };
 
 } // namespace mesh::connection
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -25,23 +25,23 @@ namespace mesh::connection {
  */
 class ST2110 : public Connection {
   public:
-    static int mesh_video_format_to_st_format(int fmt, st_frame_fmt &st_fmt);
-    static int mesh_audio_format_to_st_format(int fmt, st30_fmt &st_fmt);
-    static int mesh_audio_sampling_to_st_sampling(int sampling, st30_sampling &st_sampling);
-    static int mesh_audio_ptime_to_st_ptime(int ptime, st30_ptime &st_ptime);
+    static int mesh_video_format_to_st_format(int fmt, st_frame_fmt& st_fmt);
+    static int mesh_audio_format_to_st_format(int fmt, st30_fmt& st_fmt);
+    static int mesh_audio_sampling_to_st_sampling(int sampling, st30_sampling& st_sampling);
+    static int mesh_audio_ptime_to_st_ptime(int ptime, st30_ptime& st_ptime);
 
     static void *get_frame_data_ptr(st_frame *src);
     static void *get_frame_data_ptr(st30_frame *src);
 
-    static void get_mtl_dev_params(mtl_init_params &st_param, const std::string &dev_port,
+    static void get_mtl_dev_params(mtl_init_params& st_param, const std::string& dev_port,
                                    mtl_log_level log_level,
                                    const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
-    static mtl_handle get_mtl_device(const std::string &dev_port, mtl_log_level log_level,
+    static mtl_handle get_mtl_device(const std::string& dev_port, mtl_log_level log_level,
                                      const char local_ip_addr[MESH_IP_ADDRESS_SIZE],
-                                     int &session_id);
+                                     int& session_id);
 
     ST2110() : mtl_device(nullptr) {};
-    virtual ~ST2110(){};
+    virtual ~ST2110() {};
 
   protected:
     static int frame_available_cb(void *ptr);
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -48,8 +48,6 @@ class ST2110 : public Connection {
 
     mtl_handle mtl_device;
     std::atomic<bool> stop;
-    std::condition_variable_any cv;
-    std::mutex mx;
 };
 
 } // namespace mesh::connection
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -11,11 +11,7 @@
 #include "mesh_dp.h"
 #include "logger.h"
 
-namespace mesh
-{
-
-namespace connection
-{
+namespace mesh::connection {
 
 #define ST_APP_PAYLOAD_TYPE_ST30 (111)
 #define ST_APP_PAYLOAD_TYPE_ST20 (112)
@@ -27,8 +23,7 @@ namespace connection
  * Base abstract class of ST2110. ST2110Rx/ST2110Tx
  * inherit this class.
  */
-class ST2110 : public Connection
-{
+class ST2110 : public Connection {
   public:
     static st_frame_fmt mesh_video_format_to_st_format(int fmt);
     static st30_fmt mesh_audio_format_to_st_format(int fmt);
@@ -56,8 +51,6 @@ class ST2110 : public Connection
   private:
 };
 
-} // namespace connection
-
-} // namespace mesh
+} // namespace mesh::connection
 
 #endif // ST2110_H
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -47,6 +47,7 @@ class ST2110 : public Connection {
     mtl_handle _st;
     std::atomic<bool> _stop;
     std::condition_variable_any _cv;
+    std::mutex _mx;
 
   private:
 };
--- a/media-proxy/include/mesh/st2110.h
+++ b/media-proxy/include/mesh/st2110.h
@@ -45,11 +45,9 @@ class ST2110 : public Connection {
     static int frame_available_cb(void *ptr);
 
     mtl_handle mtl_device;
-    std::atomic<bool> _stop;
-    std::condition_variable_any _cv;
-    std::mutex _mx;
-
-  private:
+    std::atomic<bool> stop;
+    std::condition_variable_any cv;
+    std::mutex mx;
 };
 
 } // namespace mesh::connection
--- a/media-proxy/include/mesh/st2110rx.h
+++ b/media-proxy/include/mesh/st2110rx.h
@@ -11,81 +11,39 @@ namespace mesh::connection {
  * Base abstract class of ST2110Rx. ST2110_20Rx/ST2110_22Rx/ST2110_30Rx
  * inherit this class.
  */
-template <typename FRAME, typename HANDLE, typename OPS> class ST2110Rx : public ST2110 {
+template <typename FRAME, typename HANDLE, typename OPS>
+class ST2110Rx : public ST2110<FRAME, HANDLE, OPS> {
   public:
-    ST2110Rx() { _kind = Kind::receiver; }
-
-    ~ST2110Rx() {
-        shutdown(_ctx);
-        if (ops.name)
-            free((void *)ops.name);
-    }
+    ST2110Rx() { this->_kind = Kind::receiver; }
 
   protected:
-    HANDLE mtl_session = nullptr;
-    OPS ops = {0};
-    size_t transfer_size = 0;
     std::jthread frame_thread_handle;
-    context::Context _ctx = context::WithCancel(context::Background());
-
-    virtual FRAME *get_frame(HANDLE) = 0;
-    virtual int put_frame(HANDLE, FRAME *) = 0;
-    virtual HANDLE create_session(mtl_handle, OPS *) = 0;
-    virtual int close_session(HANDLE) = 0;
 
     int configure_common(context::Context& ctx, const std::string& dev_port,
-                         const MeshConfig_ST2110& cfg_st2110) {
-        int session_id = 0;
-        mtl_device = get_mtl_device(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr, session_id);
-        if (!mtl_device) {
-            log::error("Failed to get MTL device");
-            return -1;
-        }
+                         const MeshConfig_ST2110& cfg_st2110) override{
+        ST2110<FRAME, HANDLE, OPS>::configure_common(ctx, dev_port, cfg_st2110);
 
-        inet_pton(AF_INET, cfg_st2110.remote_ip_addr, ops.port.ip_addr[MTL_PORT_P]);
-        inet_pton(AF_INET, cfg_st2110.local_ip_addr, ops.port.mcast_sip_addr[MTL_PORT_P]);
-        ops.port.udp_port[MTL_PORT_P] = cfg_st2110.local_port;
-
-        strlcpy(ops.port.port[MTL_PORT_P], dev_port.c_str(), MTL_PORT_MAX_LEN);
-        ops.port.num_port = 1;
-
-        char session_name[NAME_MAX] = "";
-        snprintf(session_name, NAME_MAX, "mcm_mtl_rx_%d", session_id);
-        if (ops.name)
-            free((void *)ops.name);
-        ops.name = strdup(session_name);
-        ops.framebuff_cnt = 4;
-
-        ops.priv = this; // app handle register to lib
-        ops.notify_frame_available = frame_available_cb;
+        inet_pton(AF_INET, cfg_st2110.remote_ip_addr, this->ops.port.ip_addr[MTL_PORT_P]);
+        inet_pton(AF_INET, cfg_st2110.local_ip_addr, this->ops.port.mcast_sip_addr[MTL_PORT_P]);
+        this->ops.port.udp_port[MTL_PORT_P] = cfg_st2110.local_port;
 
         log::info("ST2110Rx: configure")
-            ("port", ops.port.port[MTL_PORT_P])
-            ("ip_addr", std::to_string(ops.port.ip_addr[MTL_PORT_P][0]) + " " +
-                        std::to_string(ops.port.ip_addr[MTL_PORT_P][1]) + " " +
-                        std::to_string(ops.port.ip_addr[MTL_PORT_P][2]) + " " +
-                        std::to_string(ops.port.ip_addr[MTL_PORT_P][3]))
-            ("mcast_sip_addr", std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][0]) + " " +
-                               std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][1]) + " " +
-                               std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][2]) + " " +
-                               std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][3]))
-            ("num_port", ops.port.num_port)
-            ("udp_port", ops.port.udp_port[MTL_PORT_P])
-            ("name", ops.name)
-            ("framebuff_cnt", ops.framebuff_cnt);
-
+            ("ip_addr", std::to_string(this->ops.port.ip_addr[MTL_PORT_P][0]) + " " +
+                        std::to_string(this->ops.port.ip_addr[MTL_PORT_P][1]) + " " +
+                        std::to_string(this->ops.port.ip_addr[MTL_PORT_P][2]) + " " +
+                        std::to_string(this->ops.port.ip_addr[MTL_PORT_P][3]))
+            ("mcast_sip_addr", std::to_string(this->ops.port.mcast_sip_addr[MTL_PORT_P][0]) + " " +
+                               std::to_string(this->ops.port.mcast_sip_addr[MTL_PORT_P][1]) + " " +
+                               std::to_string(this->ops.port.mcast_sip_addr[MTL_PORT_P][2]) + " " +
+                               std::to_string(this->ops.port.mcast_sip_addr[MTL_PORT_P][3]))
+            ("udp_port", this->ops.port.udp_port[MTL_PORT_P]);
         return 0;
     }
 
     Result on_establish(context::Context& ctx) override {
-        _ctx = context::WithCancel(ctx);
-        init_frame_available();
-
-        mtl_session = create_session(mtl_device, &ops);
-        if (!mtl_session) {
-            log::error("Failed to create session");
-            set_state(ctx, State::closed);
-            return set_result(Result::error_general_failure);
+        Result res = ST2110<FRAME, HANDLE, OPS>::on_establish(ctx);
+        if (res != Result::success) {
+            return res;
         }
 
         /* Start MTL session thread. */
@@ -93,42 +51,38 @@ template <typename FRAME, typename HANDLE, typename OPS> class ST2110Rx : public
             frame_thread_handle = std::jthread(&ST2110Rx::frame_thread, this);
         } catch (const std::system_error& e) {
             log::error("Failed to create thread");
-            set_state(ctx, State::closed);
-            return set_result(Result::error_out_of_memory);
+            this->set_state(ctx, State::closed);
+            return this->set_result(Result::error_out_of_memory);
         }
 
-        set_state(ctx, State::active);
-        return set_result(Result::success);
+        this->set_state(ctx, State::active);
+        return this->set_result(Result::success);
     }
 
     Result on_shutdown(context::Context& ctx) override {
-        _ctx.cancel();
-        notify_frame_available();
+        Result res = ST2110<FRAME, HANDLE, OPS>::on_shutdown(ctx);
+        if (res != Result::success) {
+            return res;
+        }
 
         frame_thread_handle.join();
 
-        if (mtl_session) {
-            close_session(mtl_session);
-            mtl_session = nullptr;
-        }
-        set_state(ctx, State::closed);
-        return set_result(Result::success);
+        this->set_state(ctx, State::closed);
+        return this->set_result(Result::success);
     };
-
-    virtual void on_delete(context::Context& ctx) override {}
 
   private:
     void frame_thread() {
-        while (!_ctx.cancelled()) {
+        while (!this->_ctx.cancelled()) {
             // Get full buffer from MTL
-            FRAME *frame_ptr = get_frame(mtl_session);
+            FRAME *frame_ptr = this->get_frame(this->mtl_session);
             if (frame_ptr) {
                 // Forward buffer to emulated receiver
-                transmit(_ctx, get_frame_data_ptr(frame_ptr), transfer_size);
+                this->transmit(this->_ctx, get_frame_data_ptr(frame_ptr), this->transfer_size);
                 // Return used buffer to MTL
-                put_frame(mtl_session, frame_ptr);
+                this->put_frame(this->mtl_session, frame_ptr);
             } else {
-                wait_frame_available();
+                this->wait_frame_available();
             }
         }
     }
--- a/media-proxy/include/mesh/st2110rx.h
+++ b/media-proxy/include/mesh/st2110rx.h
@@ -15,19 +15,69 @@ template <typename FRAME, typename HANDLE, typename OPS> class ST2110Rx : public
   public:
     ST2110Rx() : mtl_session(nullptr), ops{0}, transfer_size(0) { _kind = Kind::receiver; }
 
-    ~ST2110Rx() { shutdown(_ctx); }
+    ~ST2110Rx()
+    {
+        shutdown(_ctx);
+        if (ops.name)
+            free((void *)ops.name);
+    }
 
   protected:
     HANDLE mtl_session;
     OPS ops;
     size_t transfer_size;
     std::jthread frame_thread_handle;
-    context::Context _ctx;
+    context::Context _ctx = context::WithCancel(context::Background());
 
     virtual FRAME *get_frame(HANDLE) = 0;
     virtual int put_frame(HANDLE, FRAME *) = 0;
     virtual HANDLE create_session(mtl_handle, OPS *) = 0;
     virtual int close_session(HANDLE) = 0;
+
+    int configure_common(context::Context &ctx, const std::string &dev_port,
+                         const MeshConfig_ST2110 &cfg_st2110)
+    {
+        int session_id = 0;
+        mtl_device = get_mtl_device(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr, session_id);
+        if (mtl_device == nullptr) {
+            log::error("Failed to get MTL device");
+            return -1;
+        }
+
+        inet_pton(AF_INET, cfg_st2110.remote_ip_addr, ops.port.ip_addr[MTL_PORT_P]);
+        inet_pton(AF_INET, cfg_st2110.local_ip_addr, ops.port.mcast_sip_addr[MTL_PORT_P]);
+        ops.port.udp_port[MTL_PORT_P] = cfg_st2110.local_port;
+
+        strlcpy(ops.port.port[MTL_PORT_P], dev_port.c_str(), MTL_PORT_MAX_LEN);
+        ops.port.num_port = 1;
+
+        char session_name[NAME_MAX] = "";
+        snprintf(session_name, NAME_MAX, "mcm_mtl_rx_%d", session_id);
+        if (ops.name)
+            free((void *)ops.name);
+        ops.name = strdup(session_name);
+        ops.framebuff_cnt = 4;
+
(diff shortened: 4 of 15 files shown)


media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,146 @@
...
+#define ST2110TX_H
+
+#include "st2110.h"
+#include <algorithm>
+
+namespace mesh
+{
+
+namespace connection
+{
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/ed6d1ec059f95e4deb0fca4fde55b57b0c17eb15

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,146 @@
...
+
+    Result configure(context::Context &ctx, const std::string &dev_port,
+                     const MeshConfig_ST2110 &cfg_st2110, const MeshConfig_Audio &cfg_audio);
+
+  private:
+};
+
+} // namespace connection
+
+} // namespace mesh
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/ed6d1ec059f95e4deb0fca4fde55b57b0c17eb15

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,138 @@
...
+    ~ST2110Rx() { shutdown(_ctx); }
+
+  protected:
+    HANDLE _handle;
+    OPS _ops;
+    size_t _transfer_size;
+    std::jthread _frame_thread_handle;
+    context::Context _ctx;
+
+    std::function<FRAME *(HANDLE)> _get_frame_fn;
```
Sakoram, line 33: Why this isn't just a virtual function which is overwritten by concrete function in ST2110_**Rx?

Sakoram: I think that dynamic dispatch will do its magic and correct function will be called in e.g.  on_establish()

tszumski: Is it possible to overwrite virtual function with library API call ?
_get_frame_fn points to MTL API call st20p_rx_get_frame(), st22p_rx_get_frame(), st30p_rx_get_frame()

Sakoram: I think it isn't possible to do it in runtime, but do you need that? I see that you assign functions to those function pointers in constructors, so it looks like you are doing the job of c++ polymorphism by hand.

If I understand the code correctly you anyway will create objects of derived class, like ST2110_20Rx, so _get_frame_fn (the virtual method in this case) will point to the correct concrete method.

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
...
+    }
+    st_param.lcores = NULL;
+    st_param.memzone_max = 9000;
+}
+
+mtl_handle ST2110::get_mtl_handle(const std::string &dev_port, mtl_log_level log_level,
+                                  const char local_ip_addr[MESH_IP_ADDRESS_SIZE])
+{
+    static mtl_handle dev_handle;
+    static std::mutex mtx;
```
Sakoram, line 128: What is a shared resource protected by this mutex? Is it the same resource in ::configure()?

tszumski: Shared resource is mtl_handle, it can be initialized only once for whole application/process

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual void on_delete(context::Context &ctx) override {}
+
+  private:
+    void frame_thread()
+    {
+        while (!_ctx.cancelled()) {
+            // Get full buffer from MTL
+            FRAME *frame_ptr = get_frame(_handle);
+            if (!frame_ptr) { /* no frame */
+                std::mutex mx;
```
Sakoram, line 86: I'm not sure what this mutex protects, but when you create it in the local scope of an if statement each thread running this code will create its own mutex and lock only its own. This won't protect anything because multiple threads can have their own mutex locked, regardless of whether other threads have locked their mutex.

Please make the mutex at least static in this function, but ideally I would suggest putting all mutexes in the class scope rather than in the method scope.

Sakoram: Here, a local std::mutex is created within the do-while loop.
Every iteration creates a new mutex, which is passed to std::condition_variable::wait.
However, std::condition_variable requires consistent use of the same mutex across threads. Otherwise, it can lead to undefined behavior or missed notifications.

ko80: Exactly. Here you need to only define the local lock which holds the mutex. The mutex itself must be available for all other threads.

.clang-format
```diff
@@ -3,7 +3,11 @@ BasedOnStyle: LLVM
 IndentWidth: 4
 UseTab: Never
 ColumnLimit: 100
-BreakBeforeBraces: Linux
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: false
+  AfterNamespace: false
+  AfterFunction: true
```
ko80, line 10: This setting makes the following change

Before
```cpp
class MyClass {
public:
    void myFunction() {
        // function body
    }
};
```
After
```cpp
class MyClass {
public:
    void myFunction()
    {
        // function body
    }
};
```
This change is okay for `.cc` files, but it shouldn't apply to the header files.

media-proxy/include/mesh/st2110.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "conn.h"
+#include "mesh_dp.h"
+#include "logger.h"
+
+namespace mesh::connection {
+
+#define ST_APP_PAYLOAD_TYPE_ST30 (111)
+#define ST_APP_PAYLOAD_TYPE_ST20 (112)
+#define ST_APP_PAYLOAD_TYPE_ST22 (114)
```
ko80, line 18: Add inline comments referring to the SMPTE definition of that. You may include a link to documentation.

media-proxy/include/mesh/st2110.h
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * ST2110
+ *
+ * Base abstract class of ST2110. ST2110Rx/ST2110Tx
```
ko80, line 23:
```suggestion
 * Base abstract class of SPMTE ST2110-xx bridge. ST2110Rx/ST2110Tx
```

media-proxy/include/mesh/st2110.h
```diff
@@ -0,0 +1,57 @@
...
+ * inherit this class.
+ */
+class ST2110 : public Connection {
+  public:
+    static st_frame_fmt mesh_video_format_to_st_format(int fmt);
+    static st30_fmt mesh_audio_format_to_st_format(int fmt);
+    static st30_sampling mesh_audio_sampling_to_st_sampling(int sampling);
+    static st30_ptime mesh_audio_ptime_to_st_ptime(int ptime);
```
ko80, line 31: Consider removing the prefix `mesh_`. It is redundant since the static methods are defined under `namespace mesh::connection`.

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
...
+    case MESH_VIDEO_PIXEL_FORMAT_YUV444P10LE:
+        return ST_FRAME_FMT_YUV444PLANAR10LE;
+    case MESH_VIDEO_PIXEL_FORMAT_RGB8:
+        return ST_FRAME_FMT_RGB8;
+    default:
+        return ST_FRAME_FMT_YUV422PLANAR10LE;
+    }
+}
+
+st30_fmt ST2110::mesh_audio_format_to_st_format(int fmt)
```
ko80, line 23: Rework this to return an error if the format is unknown.

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
...
+        return ST31_PTIME_0_14MS;
+    case MESH_AUDIO_PACKET_TIME_0_09MS:
+        return ST31_PTIME_0_09MS;
+    default:
+        return ST30_PTIME_MAX;
+    }
+}
+
+void *ST2110::get_frame_data_ptr(st_frame *src) { return src->addr[0]; }
+void *ST2110::get_frame_data_ptr(st30_frame *src) { return src->addr; }
```
ko80, line 78: Do not inline these in `.cc` files. Expand to a normal function block OR move to the header file.

ko80: Are you sure `src` is always non NULL?

media-proxy/src/mesh/st2110_20rx.cc
```diff
@@ -0,0 +1,88 @@
...
+    static std::mutex mtx;
+    std::lock_guard<std::mutex> lock(mtx);
+
+    if (cfg_st2110.transport != MESH_CONN_TRANSPORT_ST2110_20) {
+        set_state(ctx, State::not_configured);
+        return set_result(Result::error_bad_argument);
+    }
+
+    _st = get_mtl_handle(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr);
+    if (_st == nullptr) {
```
ko80, line 34: You initialize `_st` with `0`. Why do check it to be `nullptr` here then? Avoid ambiguity.

media-proxy/src/mesh/st2110_22tx.cc
```diff
@@ -0,0 +1,88 @@
...
+    _st = get_mtl_handle(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr);
+    if (_st == nullptr) {
+        log::error("Failed to get MTL device");
+        set_state(ctx, State::not_configured);
+        return set_result(Result::error_bad_argument);
+    }
+
+    char session_name[NAME_MAX] = "";
+
+    snprintf(session_name, NAME_MAX, "mcm_tx_st22_%d", session_id++);
```
ko80, line 42: `session_id` is not thread safe. Be careful!

media-proxy/src/mesh/st2110_22tx.cc
```diff
@@ -0,0 +1,88 @@
...
+    _ops.port.udp_port[MTL_PORT_P] = cfg_st2110.remote_port;
+    strlcpy(_ops.port.port[MTL_PORT_P], dev_port.c_str(), MTL_PORT_MAX_LEN);
+    _ops.port.udp_src_port[MTL_PORT_P] = cfg_st2110.local_port;
+    _ops.port.num_port = 1;
+    _ops.port.payload_type = ST_APP_PAYLOAD_TYPE_ST22;
+    _ops.name = strdup(session_name);
+    _ops.width = cfg_video.width;
+    _ops.height = cfg_video.height;
+    _ops.fps = st_frame_rate_to_st_fps(cfg_video.fps);
+    _ops.input_fmt = mesh_video_format_to_st_format(cfg_video.pixel_format);
```
ko80, line 54: Honestly, these underscore characters are a visual noise for me. Do you really need them in the field names?

media-proxy/src/mesh/st2110_22tx.cc
```diff
@@ -0,0 +1,88 @@
...
+    _ops.device = ST_PLUGIN_DEVICE_AUTO;
+    _ops.framebuff_cnt = 4;
+    _ops.pack_type = ST22_PACK_CODESTREAM;
+    _ops.codec = ST22_CODEC_JPEGXS;
+    _ops.quality = ST22_QUALITY_MODE_SPEED;
+    _ops.codec_thread_cnt = 0;
+    _ops.codestream_size = _ops.width * _ops.height * 3 / 8;
```
ko80, line 61: Do you hardcode the pixel format here?

media-proxy/src/mesh/st2110_22tx.cc
```diff
@@ -0,0 +1,88 @@
...
+    log::info("udp_port      : %d", _ops.port.udp_port[MTL_PORT_P]);
+    log::info("udp_src_port  : %d", _ops.port.udp_src_port[MTL_PORT_P]);
+    log::info("payload_type  : %d", _ops.port.payload_type);
+    log::info("name          : %s", _ops.name);
+    log::info("width         : %d", _ops.width);
+    log::info("height        : %d", _ops.height);
+    log::info("fps           : %d", _ops.fps);
+    log::info("input_fmt     : %d", _ops.input_fmt);
+    log::info("device        : %d", _ops.device);
+    log::info("framebuff_cnt : %d", _ops.framebuff_cnt);
```
ko80, line 78: The structured logger is to avoid these tables in the log output and to make the log messages be machine-readable (when the JSON format is configured). Rework this to use the structured logging feature.

media-proxy/src/mesh/st2110_30rx.cc
```diff
@@ -0,0 +1,85 @@
+#include "st2110rx.h"
+
+namespace mesh::connection {
+
+ST2110_30Rx::ST2110_30Rx() {}
+
+ST2110_30Rx::~ST2110_30Rx() {}
```
ko80, line 7: I suggest moving these to the header file.

media-proxy/src/mesh/st2110_30rx.cc
```diff
@@ -0,0 +1,85 @@
...
+
+st30_frame *ST2110_30Rx::get_frame(st30p_rx_handle h) { return st30p_rx_get_frame(h); };
+
+int ST2110_30Rx::put_frame(st30p_rx_handle h, st30_frame *f) { return st30p_rx_put_frame(h, f); };
```
ko80, line 11: Do not inline in `.cc` files.

media-proxy/src/mesh/st2110_30rx.cc
```diff
@@ -0,0 +1,85 @@
...
+};
+
+int ST2110_30Rx::close_session(st30p_rx_handle h) { return st30p_rx_free(h); };
+
+Result ST2110_30Rx::configure(context::Context &ctx, const std::string &dev_port,
+                              const MeshConfig_ST2110 &cfg_st2110,
+                              const MeshConfig_Audio &cfg_audio)
+{
+    static int session_id;
+    static std::mutex mtx;
```
ko80, line 25: Consider name this `mx` and lock as `lk`.

media-proxy/src/mesh/st2110_30rx.cc
```diff
@@ -0,0 +1,85 @@
...
+    snprintf(session_name, NAME_MAX, "mcm_rx_st30_%d", session_id++);
+
+    inet_pton(AF_INET, cfg_st2110.remote_ip_addr, _ops.port.ip_addr[MTL_PORT_P]);
+    inet_pton(AF_INET, cfg_st2110.local_ip_addr, _ops.port.mcast_sip_addr[MTL_PORT_P]);
+    _ops.port.udp_port[MTL_PORT_P] = cfg_st2110.local_port;
+
+    strlcpy(_ops.port.port[MTL_PORT_P], dev_port.c_str(), MTL_PORT_MAX_LEN);
+    _ops.port.num_port = 1;
+    _ops.port.payload_type = ST_APP_PAYLOAD_TYPE_ST30;
+    _ops.name = strdup(session_name);
```
ko80, line 51: Can you confirm you delete the newly created string after it is not needed?

ko80: No, you never delete it. Rework this to allow deleting it appropriately.

media-proxy/src/mesh/st2110_30rx.cc
```diff
@@ -0,0 +1,85 @@
...
+    _ops.framebuff_cnt = 4;
+    _ops.fmt = mesh_audio_format_to_st_format(cfg_audio.format);
+    _ops.channel = cfg_audio.channels;
+    _ops.sampling = mesh_audio_sampling_to_st_sampling(cfg_audio.sample_rate);
```
ko80, line 55: Handle an error here.

media-proxy/src/mesh/st2110_30tx.cc
```diff
@@ -0,0 +1,82 @@
...
+int ST2110_30Tx::put_frame(st30p_tx_handle h, st30_frame *f) { return st30p_tx_put_frame(h, f); };
+
+st30p_tx_handle ST2110_30Tx::create_session(mtl_handle h, st30p_tx_ops *o)
+{
+    return st30p_tx_create(h, o);
+};
+
+int ST2110_30Tx::close_session(st30p_tx_handle h) { return st30p_tx_free(h); };
+
+Result ST2110_30Tx::configure(context::Context &ctx, const std::string &dev_port,
```
ko80, line 20: The `configure` methods seem very similar in ST2110_20, _22, and _30. Consider extracting the flow to the base class.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+#include <algorithm>
+
+namespace mesh::connection {
+
+/**
+ * ST2110
+ *
+ * Base abstract class of ST2110. ST2110_20Tx/ST2110_22Tx/ST2110_30Tx
+ * inherit this class.
+ */
```
ko80, line 14: The comment is wrong. It has been copied from the base class.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+template <typename FRAME, typename HANDLE, typename OPS> class ST2110Tx : public ST2110 {
+  public:
+    ST2110Tx()
+    {
+        _kind = Kind::transmitter;
+
+        _handle = nullptr;
+        _ops = {0};
+        _transfer_size = 0;
```
ko80, line 23: Rework these to default initializer of the fields in the class declaration block.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+
+        set_state(ctx, State::active);
+        return set_result(Result::success);
+    };
+
+    Result on_shutdown(context::Context &ctx) override
+    {
+        _ctx.cancel();
+
+        if (_handle) {
```
ko80, line 58: Consider renaming this to `session`.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+            // Get empty buffer from MTL
+            frame = get_frame(_handle);
+            if (!frame) {
+                std::unique_lock lk(_mx);
+                _cv.wait(lk, _ctx.stop_token(), [this] { return _stop.load(); });
+                _stop = false;
+                if (_ctx.cancelled()) {
+                    break;
+                }
+            }
```
ko80, line 82: Avoid using conditional variable and/or mutex in the hot path. Use atomics only.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+            close_session(_handle);
+            _handle = nullptr;
+        }
+        set_state(ctx, State::closed);
+        return set_result(Result::success);
+    };
+
+    Result on_receive(context::Context &ctx, void *ptr, uint32_t sz, uint32_t &sent) override
+    {
+        sent = std::min(_transfer_size, sz);
```
ko80, line 68: Set this variable only when you have actually sent the data. I mean, right before `return set_result(Result::success)`.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+            // Return full buffer to MTL
+            put_frame(_handle, frame);
+        } else {
+            sent = 0;
+            return set_result(Result::error_shutdown);
+        }
+        return set_result(Result::success);
+    };
+
+  private:
```
ko80, line 97:
```suggestion
```

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+};
+
+class ST2110_20Tx : public ST2110Tx<st_frame, st20p_tx_handle, st20p_tx_ops> {
+  public:
+    ST2110_20Tx();
+    ~ST2110_20Tx();
+
+    Result configure(context::Context &ctx, const std::string &dev_port,
```
ko80, line 105:
```suggestion
    Result configure(context::Context& ctx, const std::string& dev_port,
```

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+                     const MeshConfig_ST2110 &cfg_st2110, const MeshConfig_Video &cfg_video);
+
+  protected:
+    st_frame *get_frame(st20p_tx_handle h) override;
+    int put_frame(st20p_tx_handle h, st_frame *f) override;
+    st20p_tx_handle create_session(mtl_handle h, st20p_tx_ops *o) override;
+    int close_session(st20p_tx_handle h) override;
+
+  private:
```
ko80, line 114:
```suggestion
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,155 @@
...
+            _handle = nullptr;
+        }
+        set_state(ctx, State::closed);
+        return set_result(Result::success);
+    };
+
+    virtual void on_delete(context::Context &ctx) override {}
+
+  private:
+    void frame_thread()
```
ko80, line 80: This needs to be reworked to avoid using the mutex and the condition variable. Use atomics only.

.clang-format
tszumski: I would strongly suggest to use single format coding for all type of files. Do you prefer option ``.cc`` or ``.h`` ?

media-proxy/include/mesh/st2110.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/2479ca6f6d060db56009026a4c727be98472c6dc

media-proxy/include/mesh/st2110.h
```diff
@@ -0,0 +1,57 @@
...
+    static mtl_handle get_mtl_handle(const std::string &dev_port, mtl_log_level log_level,
+                                     const char local_ip_addr[MESH_IP_ADDRESS_SIZE]);
+
+    ST2110() : _st(0){};
+    virtual ~ST2110(){};
+
+  protected:
+    static int frame_available_cb(void *ptr);
+
+    mtl_handle _st;
```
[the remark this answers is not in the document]
tszumski: 1. Renamed _st to "mtl_device"
2.  Underscore prefix removed
fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/2479ca6f6d060db56009026a4c727be98472c6dc

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
+#include "st2110.h"
+
+namespace mesh::connection {
+
+st_frame_fmt ST2110::mesh_video_format_to_st_format(int fmt)
```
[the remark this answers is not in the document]
tszumski: Fixed in cbda05db27e270e513ce1c8d50c47fd4da7ba651

media-proxy/src/mesh/st2110.cc
tszumski: Fixed in cbda05db27e270e513ce1c8d50c47fd4da7ba651

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
...
+    case MESH_AUDIO_FORMAT_PCM_S16BE:
+        return ST30_FMT_PCM16;
+    case MESH_AUDIO_FORMAT_PCM_S24BE:
+        return ST30_FMT_PCM24;
+    default:
+        return ST30_FMT_MAX;
+    }
+}
+
+st30_sampling ST2110::mesh_audio_sampling_to_st_sampling(int sampling)
```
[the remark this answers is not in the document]
tszumski: Fixed in cbda05db27e270e513ce1c8d50c47fd4da7ba651

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,153 @@
...
+    case MESH_AUDIO_SAMPLE_RATE_96000:
+        return ST30_SAMPLING_96K;
+    case MESH_AUDIO_SAMPLE_RATE_44100:
+        return ST31_SAMPLING_44K;
+    default:
+        return ST30_SAMPLING_MAX;
+    }
+}
+
+st30_ptime ST2110::mesh_audio_ptime_to_st_ptime(int ptime)
```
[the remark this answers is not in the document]
tszumski: Fixed in cbda05db27e270e513ce1c8d50c47fd4da7ba651

media-proxy/src/mesh/st2110.cc
tszumski: Yes, I am sure ```src``` is always non NULL
Fixed in f4af317bd551434672e09b15a6e4ef1fcdd526cc

media-proxy/src/mesh/st2110_20rx.cc
tszumski: Initialized with ```nullptr```
Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/2479ca6f6d060db56009026a4c727be98472c6dc

media-proxy/src/mesh/st2110_22tx.cc
tszumski: Underscore removed
fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/src/mesh/st2110_30rx.cc
tszumski: Fixed in 2479ca6f6d060db56009026a4c727be98472c6dc

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+        _handle = nullptr;
+        _ops = {0};
+        _transfer_size = 0;
+    };
+    ~ST2110Tx() { shutdown(_ctx); };
+
+  protected:
+    HANDLE _handle;
+    OPS _ops;
+    uint32_t _transfer_size;
```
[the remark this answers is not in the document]
tszumski: Fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/include/mesh/st2110tx.h
tszumski: Renamed "_handle" to "mtl_session"
Fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in e69185dab06ea1983303046f28572d6baf467706

media-proxy/src/mesh/st2110_22tx.cc
tszumski: Pixel format in st22 is set based on ```cfg_video.pixel_format```

media-proxy/src/mesh/st2110_22tx.cc
tszumski: Fixed during configure() rework https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

media-proxy/src/mesh/st2110_22tx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

media-proxy/src/mesh/st2110_22tx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

media-proxy/src/mesh/st2110_22tx.cc
```diff
@@ -0,0 +1,88 @@
...
+
+    _transfer_size = st_frame_size(_ops.input_fmt, _ops.width, _ops.height, false);
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,153 @@
...
+    context::Context _ctx;
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/d6fb54f34d3b43543423c71401f5fca75f1fbdde

.clang-format
ko80: We can consider the `Before` option for `.cc` files as well.

media-proxy/src/mesh/st2110_20rx.cc
```diff
@@ -0,0 +1,65 @@
...
+                              const MeshConfig_ST2110 &cfg_st2110,
+                              const MeshConfig_Video &cfg_video)
+{
+    if (cfg_st2110.transport != MESH_CONN_TRANSPORT_ST2110_20) {
+        set_state(ctx, State::not_configured);
+        return set_result(Result::error_bad_argument);
+    }
+
+    if (configure_common(ctx, dev_port, cfg_st2110))
+        return set_result(Result::error_bad_argument);
```
ko80, line 35: State will stay `configuring` here.

media-proxy/src/mesh/st2110_20rx.cc
```diff
@@ -0,0 +1,65 @@
...
+
+    ops.port.payload_type = ST_APP_PAYLOAD_TYPE_ST20;
+    ops.width = cfg_video.width;
+    ops.height = cfg_video.height;
+    ops.fps = st_frame_rate_to_st_fps(cfg_video.fps);
+    ops.transport_fmt = ST20_FMT_YUV_422_PLANAR10LE;
+
+    if (mesh_video_format_to_st_format(cfg_video.pixel_format, ops.output_fmt))
+        return set_result(Result::error_bad_argument);
```
ko80, line 44: State will stay `configuring` here.

media-proxy/src/mesh/st2110_20rx.cc
```diff
@@ -0,0 +1,65 @@
...
+
+    ops.device = ST_PLUGIN_DEVICE_AUTO;
+
+    transfer_size = st_frame_size(ops.output_fmt, ops.width, ops.height, false);
+    if (transfer_size == 0)
+        return set_result(Result::error_bad_argument);
```
ko80, line 50: State will stay `configuring` here.

media-proxy/src/mesh/st2110_20tx.cc
```diff
@@ -0,0 +1,65 @@
...
+                              const MeshConfig_ST2110 &cfg_st2110,
+                              const MeshConfig_Video &cfg_video)
+{
+    if (cfg_st2110.transport != MESH_CONN_TRANSPORT_ST2110_20) {
+        set_state(ctx, State::not_configured);
+        return set_result(Result::error_bad_argument);
+    }
+
+    if (configure_common(ctx, dev_port, cfg_st2110))
+        return set_result(Result::error_bad_argument);
```
ko80, line 35: State will stay `configuring` here.

media-proxy/src/mesh/st2110_20tx.cc
```diff
@@ -0,0 +1,65 @@
...
+
+    ops.port.payload_type = ST_APP_PAYLOAD_TYPE_ST20;
+    ops.width = cfg_video.width;
+    ops.height = cfg_video.height;
+    ops.fps = st_frame_rate_to_st_fps(cfg_video.fps);
+    ops.transport_fmt = ST20_FMT_YUV_422_PLANAR10LE;
+
+    if (mesh_video_format_to_st_format(cfg_video.pixel_format, ops.input_fmt))
+        return set_result(Result::error_bad_argument);
```
ko80, line 44: State will stay `configuring` here.

Handle these cases appropriately everywhere in the code.

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/29e8bb061fe4f49f727744b4512c6e724b7891b3

media-proxy/src/mesh/st2110_20rx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/4db69d1907a638e19df895bab4b5914102226a85

media-proxy/src/mesh/st2110_20rx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/4db69d1907a638e19df895bab4b5914102226a85

media-proxy/src/mesh/st2110_20tx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/4db69d1907a638e19df895bab4b5914102226a85

media-proxy/src/mesh/st2110_20tx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/4db69d1907a638e19df895bab4b5914102226a85

media-proxy/include/mesh/st2110.h
tszumski: Those numbers I took from older implementation of media_proxy.
Based on my knowledge it originate from MTL implementation
https://github.com/OpenVisualCloud/Media-Transport-Library/blob/main/app/src/fmt.h
Digging a bit deeper, it looks like ```payload_type``` can be any 7 bits number.
This field should be added into  ```MeshConfig_ST2110```

media-proxy/src/mesh/st2110_30rx.cc
tszumski: not applicable after configure() rework

media-proxy/include/mesh/st2110.h
Sakoram: This is not from SMPTE, If i remember correctly those are just magic numbers chosen to fit in dynamic range of payload types numbers in RTP.
https://en.wikipedia.org/wiki/RTP_payload_formats#:~:text=Payload%20identifiers%2096%E2%80%93127%20are,assigned%20port%20is%20not%20required.

media-proxy/include/mesh/st2110tx.h
tszumski: Formatting done based on clang-format file.
If this change is required, then i can adjust clang-format and apply it to all newly created files

media-proxy/src/mesh/st2110_20rx.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/4db69d1907a638e19df895bab4b5914102226a85

media-proxy/include/mesh/st2110tx.h
ko80: I'm sorry for nitpicking, but can you rename it from `_sent` to `to_be_sent`? This name will reflect the actual meaning of the variable.

tszumski: Sure, fixed :) https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/4c084d7c275723bdf0591b0aade384ce17fb1b60

media-proxy/include/mesh/st2110tx.h
ko80: It's up to you. I'll be complaining every time seeing that kind of formatting.

.clang-format
tszumski: > We can consider the `Before` option for `.cc` files as well.

Agree

media-proxy/include/mesh/st2110tx.h
tszumski: Added rule to clang-format, fixed in:
https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/04f440a1dbe9daa11c701d65dedbcd178327c594

media-proxy/include/mesh/st2110tx.h
Sakoram: Do you suggesting to use spin locks with atomic variable instead of conditional variable? I'm opting against it. Sure, it would be faster, but it would consume a whole core while waiting. Here we want to retrieve the buffer from a ring, it is not that time-critical as it will be buffered anyway until it is sent.

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,176 @@
...
+    size_t transfer_size;
+    std::jthread frame_thread_handle;
+    context::Context _ctx = context::WithCancel(context::Background());
+
+    virtual FRAME *get_frame(HANDLE) = 0;
+    virtual int put_frame(HANDLE, FRAME *) = 0;
+    virtual HANDLE create_session(mtl_handle, OPS *) = 0;
+    virtual int close_session(HANDLE) = 0;
+
+    int configure_common(context::Context& ctx, const std::string& dev_port,
```
Sakoram, line 36: wait, why there is code in header files? Please remove all function definitions from .h and move them to .cc. leave just declarations.

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,168 @@
...
+                std::unique_lock lk(mx);
+                cv.wait(lk, _ctx.stop_token(), [this] { return stop.load(); });
+                stop = false;
+                if (_ctx.cancelled()) {
+                    return set_result(Result::error_shutdown);
+                }
+            }
+        } while (!frame);
+
+        if (frame) {
```
Sakoram, line 116: it will be always true as the while above exits only when frame != NULL

media-proxy/include/mesh/st2110tx.h
ko80: C++ introduced `.wait()`. Read this https://en.cppreference.com/w/cpp/atomic/atomic/wait

ko80: This had been a very anticipated feature for long time.

ko80: We already use it in the base connection class https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/3ade772ee4ae200c2186869215bf5e1ac5fca557/media-proxy/src/mesh/conn.cc#L201

media-proxy/include/mesh/st2110rx.h
ko80: Well, this shouldn't be that strict rule. It's quite a typical and industry-wide adopted code style. Let Tomasz decide on that.

tszumski: I wont be moving that code to ```.cc``` file simply because template implementation have to be in header file. I can only move implementation to subsequent header ```st2110rx_impl.h```and include it everywhere.

ko80: I believe, `st2110rx_impl.h` is not needed for now.

media-proxy/include/mesh/st2110tx.h
Sakoram: oh, cool. I didn't know about that one. In this case it is ok. 

I see that std::atomic::wait is even a little bit more readable.

media-proxy/include/mesh/st2110rx.h
Sakoram: I don't know. Placing code in a header file will increase compilation time because the code will be compiled in every file it is included in.

This could be handy for example for small libraries where all the code is in a header file, but I'm not sure if this is good design choose for our code-style.

ko80: I belive the guards that we have in every header file do the job
```cpp
#ifndef HEADER_FILE_H
#define HEADER_FILE_H

// Header file content

#endif // HEADER_FILE_H
```

tszumski: This is a template class, it will be compiled only if you create subclass.
Also header it is protected by macro to avoid multiple compilation 
```
#ifndef ST2110RX_H
#define ST2110RX_H
#endif
```
I dont think this is an issue

Sakoram: the #ifndef guard will ensure that it is included just ones in every translation unit, but this can be included in multiple translation unit, basically in multiple .cc files.

The argument about template class sounds convincing. I have to check it

ko80: Yeah, the guards allow including a header file multiple times in `.cc` and `.h` files ensuring it will be compiled only once.

media-proxy/include/mesh/conn.h
```diff
@@ -66,7 +66,7 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
-
+    error_shutdown,
```
ko80, line 69:
```suggestion
    error_context_cancelled,
```

media-proxy/src/mesh/conn.cc
```diff
@@ -273,6 +274,7 @@ const char * result2str(Result res)
     case Result::error_bad_argument:     return "bad argument";
     case Result::error_out_of_memory:    return "out of memory";
     case Result::error_general_failure:  return "general failure";
+    case Result::error_shutdown:         return "processing shutdown";
```
ko80, line 277:
```suggestion
    case Result::error_context_cancelled: return "context cancelled";
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * ST2110Rx
+ *
+ * Base abstract class of ST2110Rx. ST2110_20Rx/ST2110_22Rx/ST2110_30Rx
+ * inherit this class.
+ */
+template <typename FRAME, typename HANDLE, typename OPS> class ST2110Rx : public ST2110 {
+  public:
+    ST2110Rx() : mtl_session(nullptr), ops{0}, transfer_size(0) { _kind = Kind::receiver; }
```
ko80, line 16: Can you replace mtl_session, ops, and transfer_size with default initializers?

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+    virtual FRAME *get_frame(HANDLE) = 0;
+    virtual int put_frame(HANDLE, FRAME *) = 0;
+    virtual HANDLE create_session(mtl_handle, OPS *) = 0;
+    virtual int close_session(HANDLE) = 0;
+
+    int configure_common(context::Context& ctx, const std::string& dev_port,
+                         const MeshConfig_ST2110& cfg_st2110) {
+        int session_id = 0;
+        mtl_device = get_mtl_device(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr, session_id);
+        if (mtl_device == nullptr) {
```
ko80, line 40: Maybe this way?
```suggestion
        if (!mtl_device) {
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+        ops.framebuff_cnt = 4;
+
+        ops.priv = this; // app handle register to lib
+        ops.notify_frame_available = frame_available_cb;
+
+        log::info("ST2110Rx: configure")("port", ops.port.port[MTL_PORT_P])(
+            "ip_addr", std::to_string(ops.port.ip_addr[MTL_PORT_P][0]) + " " +
+                           std::to_string(ops.port.ip_addr[MTL_PORT_P][1]) + " " +
+                           std::to_string(ops.port.ip_addr[MTL_PORT_P][2]) + " " +
+                           std::to_string(ops.port.ip_addr[MTL_PORT_P][3]))(
```
ko80, line 66:
```suggestion
            "ip_addr", std::to_string(ops.port.ip_addr[MTL_PORT_P][0]) + " " +
                       std::to_string(ops.port.ip_addr[MTL_PORT_P][1]) + " " +
                       std::to_string(ops.port.ip_addr[MTL_PORT_P][2]) + " " +
                       std::to_string(ops.port.ip_addr[MTL_PORT_P][3]))(
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+            "mcast_sip_addr", std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][0]) + " " +
+                                  std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][1]) + " " +
+                                  std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][2]) + " " +
+                                  std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][3]))(
```
ko80, line 70:
```suggestion
            "mcast_sip_addr", std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][0]) + " " +
                              std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][1]) + " " +
                              std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][2]) + " " +
                              std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][3]))(
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+        if (ops.name)
+            free((void *)ops.name);
+        ops.name = strdup(session_name);
+        ops.framebuff_cnt = 4;
+
+        ops.priv = this; // app handle register to lib
+        ops.notify_frame_available = frame_available_cb;
+
+        log::info("ST2110Rx: configure")("port", ops.port.port[MTL_PORT_P])(
+            "ip_addr", std::to_string(ops.port.ip_addr[MTL_PORT_P][0]) + " " +
```
ko80, line 63: Please rework logging everywhere to start the line with the opening bracket `(` at the begining of each line. This should be applied to the logging parameters only.
```suggestion
        log::info("ST2110Rx: configure")("port", ops.port.port[MTL_PORT_P])
            ("ip_addr", std::to_string(ops.port.ip_addr[MTL_PORT_P][0]) + " " +
```

media-proxy/include/mesh/st2110rx.h
Sakoram: if it was just normal executable code ( I'm still not sure about template class) it would be compiled multiple times. Every source file is preprocessed separately ( so every include in sticked into every source file ) the result of this is translation unit which is then compiled. If we add code into header file it will be copied to every source file at preprocessing and then compiled multiple times. 

`#define HEADER_FILE_H` works only in single translation unit it is not propagated to all, so `#ifndef HEADER_FILE_H` ensures that we won't include single header file multiple times to a single source file, but it has to be included to every source file that actually needs it, and it will.

ko80: Oh, I didn't think about this from that perspective, so may be you're right. But still, there shouldn't be any significant latency introduced by this code style in this project. Let's keep it as it is.

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,182 @@
...
+}
+
+int ST2110::frame_available_cb(void *ptr) {
+    auto _this = static_cast<ST2110 *>(ptr);
+    if (!_this) {
+        return -1;
+    }
+
+    _this->stop = true;
+    _this->stop.notify_all();
```
ko80, line 116:
```suggestion
    _this->frame_available = true;
    _this->frame_available.notify_one();
```

media-proxy/include/mesh/st2110.h
```diff
@@ -0,0 +1,55 @@
...
+                                     int& session_id);
+
+    ST2110() : mtl_device(nullptr) {};
+    virtual ~ST2110() {};
+
+  protected:
+    static int frame_available_cb(void *ptr);
+
+    mtl_handle mtl_device;
+    std::atomic<bool> stop;
```
ko80, line 50:
```suggestion
    std::atomic<bool> frame_available;
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+                                  std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][3]))(
+            "num_port", ops.port.num_port)("udp_port", ops.port.udp_port[MTL_PORT_P])(
+            "name", ops.name)("framebuff_cnt", ops.framebuff_cnt);
+
+        return 0;
+    }
+
+    Result on_establish(context::Context& ctx) override {
+        _ctx = context::WithCancel(ctx);
+        stop = false;
```
ko80, line 79:
```suggestion
        frame_available = false;
```

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        set_state(ctx, State::active);
+        return set_result(Result::success);
+    }
+
+    Result on_shutdown(context::Context& ctx) override {
+        _ctx.cancel();
+        stop = true;
+        stop.notify_all();
```
ko80, line 104:
```suggestion
        frame_available.store(true, std::memory_order_release);
        frame_available.notify_one();
```

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,165 @@
...
+            "num_port", ops.port.num_port)("udp_port", ops.port.udp_port[MTL_PORT_P])(
+            "udp_src_port", ops.port.udp_src_port[MTL_PORT_P])("name", ops.name)("framebuff_cnt",
+                                                                                 ops.framebuff_cnt);
+
+        return 0;
+    }
+
+    Result on_establish(context::Context& ctx) override {
+        _ctx = context::WithCancel(ctx);
+        stop = false;
```
ko80, line 74:
```suggestion
        frame_available = false;
```

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        set_state(ctx, State::active);
+        return set_result(Result::success);
+    };
+
+    Result on_shutdown(context::Context& ctx) override {
+        _ctx.cancel();
+        stop = true;
+        stop.notify_all();
```
ko80, line 90: Can you refactor this to be `frame_available` and move this to the base class? Same functionality is in Rx.

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,174 @@
...
+            FRAME *frame_ptr = get_frame(mtl_session);
+            if (!frame_ptr) { /* no frame */
+                stop.wait(false);
+                stop = false;
+                continue;
+            }
+            // Forward buffer to emulated receiver
+            transmit(_ctx, get_frame_data_ptr(frame_ptr), transfer_size);
+            // Return used buffer to MTL
+            put_frame(mtl_session, frame_ptr);
```
ko80, line 131:
```suggestion
            if (frame_ptr) {
                // Forward buffer to emulated receiver
                transmit(_ctx, get_frame_data_ptr(frame_ptr), transfer_size);
                // Return used buffer to MTL
                put_frame(mtl_session, frame_ptr);
            } else {
                frame_available.wait(true, std::memory_order_acquire);
            }
```

media-proxy/include/mesh/st2110rx.h
Sakoram: ok, linker is smart. The linker eliminates duplicate template instantiations is a result of ODR merging, implemented via mechanisms like weak symbols or COMDAT folding.

It was nice to discuss. I think this code in header is ok.

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed

media-proxy/include/mesh/conn.h
tszumski: Fixed in 7b57d140d1ff8be6328bad2c1e5c18c53ad97bae

media-proxy/src/mesh/conn.cc
tszumski: Fixed in 7b57d140d1ff8be6328bad2c1e5c18c53ad97bae

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/92513fe839a25353cd6ab4ced946b0ab5ed474c9

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/6fd17e224f7feb7867c86d140b47544157cba8ff

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/92513fe839a25353cd6ab4ced946b0ab5ed474c9

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/92513fe839a25353cd6ab4ced946b0ab5ed474c9

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/92513fe839a25353cd6ab4ced946b0ab5ed474c9

media-proxy/src/mesh/st2110.cc
tszumski: Fixed in 133a6efcdea8895252fd1a825a2ae57a7f31cb01

media-proxy/include/mesh/st2110.h
tszumski: Fixed in 133a6efcdea8895252fd1a825a2ae57a7f31cb01

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in 133a6efcdea8895252fd1a825a2ae57a7f31cb01

media-proxy/include/mesh/st2110tx.h
tszumski: Fixed in 133a6efcdea8895252fd1a825a2ae57a7f31cb01

media-proxy/include/mesh/st2110rx.h
tszumski: Fixed in 133a6efcdea8895252fd1a825a2ae57a7f31cb01

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,166 @@
...
+    virtual int close_session(HANDLE) = 0;
+
+    int configure_common(context::Context& ctx, const std::string& dev_port,
+                         const MeshConfig_ST2110& cfg_st2110) {
+        int session_id = 0;
+        mtl_device = get_mtl_device(dev_port, MTL_LOG_LEVEL_CRIT, cfg_st2110.local_ip_addr, session_id);
+        if (!mtl_device) {
+            log::error("Failed to get MTL device");
+            return -1;
+        }
```
ko80, line 42: Can you move this to the base class? You could make this `configure_common()` method virtual, override it in derived classes, and call the inherited method where needed.

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,175 @@
...
+ * ST2110Rx
+ *
+ * Base abstract class of ST2110Rx. ST2110_20Rx/ST2110_22Rx/ST2110_30Rx
+ * inherit this class.
+ */
+template <typename FRAME, typename HANDLE, typename OPS> class ST2110Rx : public ST2110 {
+  public:
+    ST2110Rx() { _kind = Kind::receiver; }
+
+    ~ST2110Rx() {
```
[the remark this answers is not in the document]
tszumski: This require making base class as template. 
I assume significant amount of time is required, 2 levels of template class might not be too nice to read

media-proxy/include/mesh/st2110tx.h
ko80: I suggest you extract as much as you can from `configure_common()` methods in derived classes to the base class method.

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,175 @@
...
+                               std::to_string(ops.port.mcast_sip_addr[MTL_PORT_P][3]))
+            ("num_port", ops.port.num_port)
+            ("udp_port", ops.port.udp_port[MTL_PORT_P])
+            ("name", ops.name)
+            ("framebuff_cnt", ops.framebuff_cnt);
+
+        return 0;
+    }
+
+    Result on_establish(context::Context& ctx) override {
```
ko80, line 80: Can you extract the common part to a virtual method in the base class and then override it?

media-proxy/include/mesh/st2110rx.h
```diff
@@ -0,0 +1,175 @@
...
+            return set_result(Result::error_out_of_memory);
+        }
+
+        set_state(ctx, State::active);
+        return set_result(Result::success);
+    }
+
+    Result on_shutdown(context::Context& ctx) override {
+        _ctx.cancel();
+        notify_frame_available();
```
ko80, line 106: I suppose this part can also be extracted to a virtual method in the base class.

media-proxy/include/mesh/st2110rx.h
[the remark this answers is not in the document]
ko80: Okay, I get it. Let's keep it as it is for now.

media-proxy/src/mesh/st2110_30tx.cc
```diff
@@ -0,0 +1,60 @@
...
+};
+
+int ST2110_30Tx::close_session(st30p_tx_handle h) {
+    return st30p_tx_free(h);
+};
+
+Result ST2110_30Tx::configure(context::Context& ctx, const std::string& dev_port,
+                              const MeshConfig_ST2110& cfg_st2110,
+                              const MeshConfig_Audio& cfg_audio) {
+    set_state(ctx, State::not_configured);
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/261/commits/ca49db16b741b8e4246ad5457a80523147a70698

media-proxy/include/mesh/st2110tx.h
tszumski: Rework base class as template required

media-proxy/include/mesh/st2110rx.h
tszumski: Rework base class as template required

media-proxy/include/mesh/st2110rx.h
tszumski: Rework base class as template required

media-proxy/include/mesh/st2110tx.h
```diff
@@ -0,0 +1,166 @@
...
+        do {
+            // Get empty buffer from MTL
+            frame = get_frame(mtl_session);
+            if (!frame) {
+                wait_frame_available();
+                if (_ctx.cancelled()) {
+                    return set_result(Result::error_context_cancelled);
+                }
+            }
+        } while (!frame);
```
ko80, line 116:
```suggestion
        FRAME *frame;
        for (;;) {
            if (ctx.cancelled() || _ctx.cancelled())
                return set_result(Result::error_context_cancelled);
            
            // Get empty buffer from MTL
            frame = get_frame(mtl_session);
            if (frame)
                break;
            
            wait_frame_available();
        }
```

media-proxy/src/mesh/st2110.cc
```diff
@@ -0,0 +1,206 @@
...
+}
+
+mtl_handle ST2110::get_mtl_device(const std::string& dev_port, mtl_log_level log_level,
+                                  const char local_ip_addr[MESH_IP_ADDRESS_SIZE], int& session_id) {
+    static mtl_handle dev_handle;
+    static int _session_id;
+    static std::mutex mtx;
+    std::lock_guard<std::mutex> lock(mtx);
+
+    session_id = _session_id++;
```
ko80, line 182: Having this value incremented is not a robust approach to generate an id. I don't like it.

Check if you can use `generate_uuid_v4()` and change the session id type to std::string. See usage example https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/c2f586c7825931470f2074f98bb2a0baed4d2c1e/media-proxy/src/mesh/manager_local.cc#L31

Incremental ids are fine for this PR with the initial implementation, but must be reworked (if possible) in the nearest future to make that production grade.

ko80 (approved):
Looks good to me. Excellent job, Tomasz!

PR #264, opened 2024-11-22
Adding setLogLevel feature

moleksy:
* Features:
 *  - Use `mesh::log::setLogLevel(mesh::log::Level)` to dynamically adjust log filtering.
 *  - Supported log levels: `info`, `warn`, `error`, `debug`, `fatal`.
 *  - Messages below the set log level will be ignored.

### DIFF
--- a/media-proxy/src/mesh/logger.cc
+++ b/media-proxy/src/mesh/logger.cc
@@ -11,6 +11,8 @@
 
 namespace mesh::log {
 
+Level currentLogLevel = Level::info;
+
 void StandardFormatter::formatMessage(std::ostringstream& ostream, Level level,
                                       const char *format, va_list args)
 {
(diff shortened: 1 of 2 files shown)


moleksy:
How it works:
image

media-proxy/src/mesh/logger.cc
```diff
@@ -113,21 +115,24 @@ void setFormatter(std::unique_ptr<Formatter> new_formatter)
     formatter = std::move(new_formatter);
 }
 
+void setLogLevel(Level level) {
+    currentLogLevel = level;
+}
```
ko80, line 120: This function is not thread-safe. Since there is no meaning in changing the log level at runtime, I suggest to leave this function as it is but say in the comment block that this is to be called only once the logger is created in the app.

media-proxy/src/mesh/logger.cc
```diff
@@ -113,21 +115,24 @@ void setFormatter(std::unique_ptr<Formatter> new_formatter)
...
 }
 
-Logger::~Logger()
-{
-    if (formatter) 
-        formatter->formatAfter(ostream);
-
-    if (!ostream.str().empty())
+Logger::~Logger() {
+    if (!ostream.str().empty() && level >= currentLogLevel) {
```
ko80, line 131:
```suggestion
    if (level >= currentLogLevel && !ostream.str().empty()) {
```

media-proxy/src/mesh/logger.cc
moleksy: Actually there was a meaning for me - for different tests I wanted to have control over what is visible where, I'll make it thread safe.

ko80: If you make it thread safe, you'll make `currentLogLevel` thread safe. It will add latency to printing messages. Please don't.

PR #269, opened 2024-11-27
Fix memory leak in Context class

tszumski:
Fix memory leak in Context class

### DIFF
--- a/media-proxy/src/mesh/concurrency.cc
+++ b/media-proxy/src/mesh/concurrency.cc
@@ -14,7 +14,7 @@ namespace context {
 Context::Context() : ss(std::stop_source()),
                      parent(nullptr),
                      cb(nullptr),
-                     ch(new thread::Channel<bool>(1)),
+                     ch(nullptr),
                      timeout_ms(std::chrono::milliseconds(0))
 {
 }
@@ -55,9 +55,11 @@ Context& Context::operator=(Context&& other) noexcept {
         ss = std::stop_source();
         parent = other.parent;
 
-        cb = std::make_unique<std::stop_callback<std::function<void()>>>(
-            parent->ss.get_token(), [this] { cancel(); }
-        );
+        if(parent){
+            cb = std::make_unique<std::stop_callback<std::function<void()>>>(
+                parent->ss.get_token(), [this] { cancel(); }
+            );
+        }
 
         timeout_ms = other.timeout_ms;
 
@@ -69,6 +71,10 @@ Context& Context::operator=(Context&& other) noexcept {
                 cv.wait_for(lk, ss.get_token(), timeout_ms, [] { return false; });
                 cancel();
             });
+        }
+
+        if (ch) {
+            delete ch;
         }
 
         ch = other.ch;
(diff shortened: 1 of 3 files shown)


media-proxy/src/mesh/concurrency.cc
```diff
@@ -55,9 +55,11 @@ Context& Context::operator=(Context&& other) noexcept {
         ss = std::stop_source();
         parent = other.parent;
 
-        cb = std::make_unique<std::stop_callback<std::function<void()>>>(
-            parent->ss.get_token(), [this] { cancel(); }
-        );
+        if(parent){
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/269/commits/a4cb2a96c2db986e12fc249f66e0654509fcb401

media-proxy/src/mesh/concurrency.cc
```diff
@@ -14,7 +14,7 @@ namespace context {
 Context::Context() : ss(std::stop_source()),
                      parent(nullptr),
                      cb(nullptr),
-                     ch(new thread::Channel<bool>(1)),
+                     ch(nullptr),
```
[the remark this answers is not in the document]
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/269/commits/a4cb2a96c2db986e12fc249f66e0654509fcb401

media-proxy/tests/mesh_context_tests.cc
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ * Test designed to detect memory leaks in the Context class.
+ *
+ * Run the test with Valgrind to detect memory leaks.
+ * valgrind --leak-check=full ./media_proxy_unit_tests --gtest_filter=*Context*
+ */
+TEST(MeshContext, constructor) {
+    auto ctx = context::WithCancel(mesh::context::Background());
```
ko80, line 20:
```suggestion
    auto ctx = context::WithCancel(context::Background());
```

media-proxy/tests/mesh_context_tests.cc
```diff
@@ -0,0 +1,51 @@
...
+    test_class t;
+    t.init(ctx);
+    {
+        auto ctx2 = context::WithCancel(ctx);
+        t.init(ctx2);
+        mesh::thread::Sleep(ctx2, std::chrono::milliseconds(10));
```
ko80, line 26: I think this test case was supposed to verify the `context` implementation only. Maybe it's worth replacing `thread::Sleep` with `std::chrono::Sleep`?

ko80: But in general, I'm okay with using `thread::Sleep`.

tszumski: Lets keep it as is

media-proxy/tests/mesh_context_tests.cc
tszumski: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/269/commits/c89d2973e0e8b88369250f85b1524cb66d0ae875

ko80 (approved):
Looks good. `mesh::` is redundant.

PR #286, opened 2024-12-05
Add RDMA Connection Implementation

moleksy:
- Introduced `conn_rdma` as a base class for RDMA connection handling.
- Added `conn_rdma_rx` and `conn_rdma_tx` for specialized Rx and Tx operations:
  - `conn_rdma_rx`: Handles RDMA reception logic with buffer management.
  - `conn_rdma_tx`: Handles RDMA transmission logic with buffer management.

**Key Changes**:
- New headers:
  - `conn_rdma.h`
  - `conn_rdma_rx.h`
  - `conn_rdma_tx.h`
- Implementation files:
  - `conn_rdma.cc`
  - `conn_rdma_rx.cc`
  - `conn_rdma_tx.cc`
- Unit tests for RDMA:
  - Added `conn_rdma_rx_tests.cc` and `conn_rdma_tx_tests.cc` to validate Rx/Tx functionalities.
  - Introduced mocks for RDMA (`conn_rdma_test_mocks.cc` and `.h`).

**Enhancements**:
- Updated `CMakeLists.txt` to include new tests and sources.
- Modified `conn.h` to align with RDMA connection structure.
- Adjusted `libfabric_ep.c` to integrate RDMA-specific operations.

**Test Coverage**:
- Verified RDMA Rx/Tx behavior through unit tests.
- Introduced `conn_rdma_real_ep_test` for real endpoint integration testing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,13 +35,23 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 add_subdirectory(${SDK_DIR})
 add_subdirectory(${MP_DIR})
 
+
+# Fetch google test dependencies
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/35d0c365609296fa4730d62057c487e3cfa030ff.zip
+)
+FetchContent_MakeAvailable(googletest)
+
+target_include_directories(media_proxy_lib PUBLIC ${SDK_INCLUDE_DIR})
 option(BUILD_UNIT_TESTS "Build and enable unit tests" ON)
 if (BUILD_UNIT_TESTS)
     enable_testing()
     add_subdirectory(${MP_TESTS_DIR})
     add_subdirectory(${SDK_TESTS_DIR})
     add_subdirectory(${TESTS_VAL_DIR})
+    target_link_libraries(media_proxy_lib PUBLIC mcm_dp gtest gmock gtest_main)
+else()
+    target_link_libraries(media_proxy_lib PUBLIC mcm_dp)
 endif()
-
-target_include_directories(media_proxy_lib PUBLIC ${SDK_INCLUDE_DIR})
-target_link_libraries(media_proxy_lib PUBLIC mcm_dp)
--- a/build.sh
+++ b/build.sh
@@ -35,7 +35,7 @@ as_root ln -s /usr/lib64/libbpf.so.1 /usr/lib/x86_64-linux-gnu/libbpf.so.1 2>/de
 as_root ldconfig
 
 # Run unit tests
-export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib"
+export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib64"
 ctest --output-on-failure --test-dir "${MCM_BUILD_DIR}" -V
 
 ln -sf "${MCM_BUILD_DIR}" "${SCRIPT_DIR}/build"
--- a/build.sh
+++ b/build.sh
@@ -35,9 +35,9 @@ as_root ln -s /usr/lib64/libbpf.so.1 /usr/lib/x86_64-linux-gnu/libbpf.so.1 2>/de
 as_root ldconfig
 
 # Run unit tests
-export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib64"
-"${MCM_BUILD_DIR}/bin/sdk_unit_tests"
-"${MCM_BUILD_DIR}/bin/media_proxy_unit_tests"
+export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib"
+ctest --output-on-failure --test-dir "${MCM_BUILD_DIR}" -V
+
 ln -sf "${MCM_BUILD_DIR}" "${SCRIPT_DIR}/build"
 
 log_info "Build Succeeded"
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -71,7 +71,6 @@ enum class Result {
     error_initialization_failed,
     error_memory_registration_failed,
     error_thread_creation_failed,
-    error_operation_cancelled,
     error_no_buffer,
     error_timeout,
     error_context_cancelled,
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -72,8 +72,6 @@ enum class Result {
     error_memory_registration_failed,
     error_thread_creation_failed,
     error_operation_cancelled,
-    error_buffer_overflow,
-    error_buffer_underflow,
     error_no_buffer,
     error_timeout,
     error_context_cancelled,
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -67,13 +67,13 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
+    error_context_cancelled,
     error_already_initialized,
     error_initialization_failed,
     error_memory_registration_failed,
     error_thread_creation_failed,
     error_no_buffer,
     error_timeout,
-    error_context_cancelled,
     
     // TODO: more error codes to be added...
 };
--- a/media-proxy/include/mesh/conn.h
+++ b/media-proxy/include/mesh/conn.h
@@ -67,8 +67,17 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
+    error_already_initialized,
+    error_initialization_failed,
+    error_memory_registration_failed,
+    error_thread_creation_failed,
+    error_operation_cancelled,
+    error_buffer_overflow,
+    error_buffer_underflow,
+    error_no_buffer,
+    error_timeout,
     error_context_cancelled,
-   
+    
     // TODO: more error codes to be added...
 };
 
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -42,7 +42,8 @@ public:
     Rdma();
     virtual ~Rdma();
     static void
-    deinit_rdma_if_needed(libfabric_ctx *mDevHandle); // Deinitialize RDMA if no active connections
+    // Deinitialize RDMA if no active connections
+    deinit_rdma_if_needed(libfabric_ctx *m_dev_handle);
 
 // Used only for Unit tests, provides access to protected members
 #ifdef UNIT_TESTS_ENABLED
@@ -78,7 +79,7 @@ protected:
     void handle_error(context::Context& ctx, const char *step);
 
     // RDMA-specific members
-    libfabric_ctx *mDevHandle;                  // RDMA device handle
+    libfabric_ctx *m_dev_handle;                // RDMA device handle
     ep_ctx_t *ep_ctx;                           // RDMA endpoint context
     ep_cfg_t ep_cfg;                            // RDMA endpoint configuration
     size_t trx_sz;                              // Data transfer size
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -1,0 +1,128 @@
+#ifndef CONN_RDMA_H
+#define CONN_RDMA_H
+
+#include "logger.h"
+#include "mesh/conn.h"
+#include "concurrency.h"
+#include "libfabric_ep.h"
+#include "libfabric_mr.h"
+#include "libfabric_cq.h"
+#include "libfabric_dev.h"
+#include "mcm_dp.h"
+#include <mutex>
+#include <cstring>
+#include <cstddef>
+#include <atomic>
+#include <queue>
+
+#ifndef RDMA_DEFAULT_TIMEOUT
+#define RDMA_DEFAULT_TIMEOUT 1
+#endif
+#ifndef MAX_BUFFER_SIZE
+#define MAX_BUFFER_SIZE 1 << 30
+#endif
+#ifndef CQ_BATCH_SIZE
+#define CQ_BATCH_SIZE 64
+#endif
+#ifndef PAGE_SIZE
+#define PAGE_SIZE 4096
+#endif
+
+namespace mesh {
+
+namespace connection {
+
+/**
+ * Rdma
+ *
+ * Base class for RDMA connections, derived from the `Connection` class.
+ * Provides common RDMA-related functionality and acts as a foundation
+ * for specialized RDMA Tx and Rx classes.
+ */
+class Rdma : public Connection {
+  public:
+    Rdma();
+    virtual ~Rdma();
+
+  //Used only for Unit tests, provides access to protected members
+  #ifdef UNIT_TESTS_ENABLED
+      size_t get_buffer_queue_size() const { return buffer_queue.size(); }
+      bool is_buffer_queue_empty() const { return buffer_queue.empty(); }
+      Kind get_kind() const { return _kind; }
+      void* get_buffer_block() const { return buffer_block; }
+  #endif
+
+    // Queue synchronization
+    void init_buf_available();
+    void notify_buf_available();
+    void wait_buf_available();
+
+  protected:
+    // Configure the RDMA session
+    virtual Result configure(context::Context& ctx, const mcm_conn_param& request,
+                             const std::string& dev_port, libfabric_ctx *& dev_handle, Kind kind,
+                             direction dir);
+
+    // Overrides from Connection
+    virtual Result on_establish(context::Context& ctx) override;
+    virtual void on_delete(context::Context& ctx) override;
+    virtual Result on_shutdown(context::Context& ctx) override;
+
+    // RDMA-specific methods
+
+    // Configure RDMA endpoint
+    Result configure_endpoint(context::Context& ctx);
+
+    // Cleanup RDMA resources
+    Result cleanup_resources(context::Context& ctx);
+
+    // Error handler for logging and recovery
+    void handle_error(context::Context& ctx, const char *step);
+
+    // RDMA-specific members
+    libfabric_ctx* mDevHandle; // RDMA device handle
+    ep_ctx_t* ep_ctx;          // RDMA endpoint context
+    ep_cfg_t ep_cfg;           // RDMA endpoint configuration
+    size_t trx_sz;             // Data transfer size
+    bool init;                 // Indicates if RDMA is initialized
+    void* buffer_block;        // Pointer to the allocated buffer block
+    int queue_size;            // Number of buffers in the queue
+
+    // Queue for managing buffers
+    std::queue<void*> buffer_queue; // Queue holding available buffers
+    std::mutex queue_mutex;         // Mutex for buffer queue synchronization
+    std::condition_variable_any queue_cv; // Condition variable for buffer availability
+
+    // // RDMA thread logic
+    // virtual void process_buffers_thread(context::Context& ctx) = 0;
+    // virtual void rdma_cq_thread(context::Context& ctx) = 0;
+    virtual Result start_threads(context::Context& ctx) { return Result::success; }
+
+    Result init_queue_with_elements(size_t capacity, size_t trx_sz);
+    Result add_to_queue(void *element);
+    Result consume_from_queue(context::Context& ctx, void **element);
+    void cleanup_queue();
+
+    std::jthread handle_process_buffers_thread; // Thread for processing buffers
+    std::jthread handle_rdma_cq_thread;         // Thread for completion queue operations
+    context::Context process_buffers_thread_ctx; // Context for buffer processing thread
+    context::Context rdma_cq_thread_ctx;        // Context for completion queue thread
+
+    std::mutex cq_mutex;              // Mutex for completion queue synchronization
+    std::condition_variable_any cq_cv; // Condition variable for completion queue events
+    bool event_ready = false;         // Indicates if an event is ready in the CQ
+
+    void notify_cq_event();
+    void shutdown_rdma(context::Context& ctx);
+
+    //Helper functions
+    std::string kind_to_string(Kind kind);
+
+    std::atomic<bool> buf_available; // Indicates buffer availability in the queue
+};
+
+} // namespace connection
+
+} // namespace mesh
+
+#endif // CONN_RDMA_H
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -58,7 +58,7 @@ class Rdma : public Connection {
   protected:
     // Configure the RDMA session
     virtual Result configure(context::Context& ctx, const mcm_conn_param& request,
-                             const std::string& dev_port, libfabric_ctx *& dev_handle);
+                             libfabric_ctx *& dev_handle);
 
     // Overrides from Connection
     virtual Result on_establish(context::Context& ctx) override;
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -41,6 +41,7 @@ class Rdma : public Connection {
   public:
     Rdma();
     virtual ~Rdma();
+    static void deinit_rdma_if_needed(libfabric_ctx *mDevHandle); // Deinitialize RDMA if no active connections
 
   //Used only for Unit tests, provides access to protected members
   #ifdef UNIT_TESTS_ENABLED
@@ -76,14 +77,15 @@ class Rdma : public Connection {
     void handle_error(context::Context& ctx, const char *step);
 
     // RDMA-specific members
-    libfabric_ctx* mDevHandle; // RDMA device handle
-    ep_ctx_t* ep_ctx;          // RDMA endpoint context
-    ep_cfg_t ep_cfg;           // RDMA endpoint configuration
-    size_t trx_sz;             // Data transfer size
-    bool init;                 // Indicates if RDMA is initialized
-    void* buffer_block;        // Pointer to the allocated buffer block
-    int queue_size;            // Number of buffers in the queue
-    direction dir;             // Data transfer direction
+    libfabric_ctx *mDevHandle;                  // RDMA device handle
+    ep_ctx_t *ep_ctx;                           // RDMA endpoint context
+    ep_cfg_t ep_cfg;                            // RDMA endpoint configuration
+    size_t trx_sz;                              // Data transfer size
+    bool init;                                  // Indicates if RDMA is initialized
+    void *buffer_block;                         // Pointer to the allocated buffer block
+    int queue_size;                             // Number of buffers in the queue
+    direction dir;                              // Data transfer direction
+    static std::atomic<int> active_connections; // Number of active RDMA connections
 
     // Queue for managing buffers
     std::queue<void*> buffer_queue; // Queue holding available buffers
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -41,9 +41,8 @@ class Rdma : public Connection {
 public:
     Rdma();
     virtual ~Rdma();
-    static void
     // Deinitialize RDMA if no active connections
-    deinit_rdma_if_needed(libfabric_ctx *m_dev_handle);
+    static void deinit_rdma_if_needed(libfabric_ctx *m_dev_handle);
 
 // Used only for Unit tests, provides access to protected members
 #ifdef UNIT_TESTS_ENABLED
--- a/media-proxy/include/mesh/conn_rdma.h
+++ b/media-proxy/include/mesh/conn_rdma.h
@@ -38,24 +38,25 @@ namespace mesh::connection {
  * for specialized RDMA Tx and Rx classes.
  */
 class Rdma : public Connection {
-  public:
+public:
     Rdma();
     virtual ~Rdma();
-    static void deinit_rdma_if_needed(libfabric_ctx *mDevHandle); // Deinitialize RDMA if no active connections
+    static void
+    deinit_rdma_if_needed(libfabric_ctx *mDevHandle); // Deinitialize RDMA if no active connections
 
-  //Used only for Unit tests, provides access to protected members
-  #ifdef UNIT_TESTS_ENABLED
-      size_t get_buffer_queue_size() const { return buffer_queue.size(); }
-      bool is_buffer_queue_empty() const { return buffer_queue.empty(); }
-      void* get_buffer_block() const { return buffer_block; }
-  #endif
+// Used only for Unit tests, provides access to protected members
+#ifdef UNIT_TESTS_ENABLED
+    size_t get_buffer_queue_size() const { return buffer_queue.size(); }
+    bool is_buffer_queue_empty() const { return buffer_queue.empty(); }
+    void *get_buffer_block() const { return buffer_block; }
+#endif
 
     // Queue synchronization
     void init_buf_available();
     void notify_buf_available();
     void wait_buf_available();
 
-  protected:
+protected:
     // Configure the RDMA session
     virtual Result configure(context::Context& ctx, const mcm_conn_param& request,
                              libfabric_ctx *& dev_handle);
@@ -88,8 +89,8 @@ class Rdma : public Connection {
     static std::atomic<int> active_connections; // Number of active RDMA connections
 
     // Queue for managing buffers
-    std::queue<void*> buffer_queue; // Queue holding available buffers
-    std::mutex queue_mutex;         // Mutex for buffer queue synchronization
+    std::queue<void *> buffer_queue;      // Queue holding available buffers
+    std::mutex queue_mutex;               // Mutex for buffer queue synchronization
     std::condition_variable_any queue_cv; // Condition variable for buffer availability
 
     // // RDMA thread logic
@@ -100,17 +101,17 @@ class Rdma : public Connection {
     Result consume_from_queue(context::Context& ctx, void **element);
     void cleanup_queue();
 
-    std::jthread handle_process_buffers_thread; // Thread for processing buffers
-    std::jthread handle_rdma_cq_thread;         // Thread for completion queue operations
+    std::jthread handle_process_buffers_thread;  // Thread for processing buffers
+    std::jthread handle_rdma_cq_thread;          // Thread for completion queue operations
     context::Context process_buffers_thread_ctx; // Context for buffer processing thread
-    context::Context rdma_cq_thread_ctx;        // Context for completion queue thread
+    context::Context rdma_cq_thread_ctx;         // Context for completion queue thread
 
-    std::mutex cq_mutex;              // Mutex for completion queue synchronization
+    std::mutex cq_mutex;               // Mutex for completion queue synchronization
     std::condition_variable_any cq_cv; // Condition variable for completion queue events
-    bool event_ready = false;         // Indicates if an event is ready in the CQ
+    bool event_ready = false;          // Indicates if an event is ready in the CQ
 
     void notify_cq_event();
-  
+
     std::atomic<bool> buf_available; // Indicates buffer availability in the queue
 };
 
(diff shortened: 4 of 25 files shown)


build.sh
```diff
@@ -35,9 +35,9 @@ as_root ln -s /usr/lib64/libbpf.so.1 /usr/lib/x86_64-linux-gnu/libbpf.so.1 2>/de
 as_root ldconfig
 
 # Run unit tests
-export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib64"
-"${MCM_BUILD_DIR}/bin/sdk_unit_tests"
-"${MCM_BUILD_DIR}/bin/media_proxy_unit_tests"
+export LD_LIBRARY_PATH="${PREFIX_DIR}/usr/local/lib:/usr/local/lib"
+ctest --output-on-failure --test-dir "${MCM_BUILD_DIR}" -V
+
```
ko80, line 40: Please verify what you're changing here

moleksy: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/316ac7128e8def77fd3380b6459edd0884ea023c

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+    int queue_size;            // Number of buffers in the queue
+
+    // Queue for managing buffers
+    std::queue<void*> buffer_queue; // Queue holding available buffers
+    std::mutex queue_mutex;         // Mutex for buffer queue synchronization
+    std::condition_variable_any queue_cv; // Condition variable for buffer availability
+
+    // // RDMA thread logic
+    // virtual void process_buffers_thread(context::Context& ctx) = 0;
+    // virtual void rdma_cq_thread(context::Context& ctx) = 0;
```
tszumski, line 98: Do you plan to use the commented methods? If not please remove them

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+    virtual Result start_threads(context::Context& ctx) { return Result::success; }
```
tszumski, line 99: Please make ```start_threads()``` pure virtual 
```
virtual Result start_threads(context::Context& ctx) = 0;
```

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+
+    // Queue synchronization
+    void init_buf_available();
+    void notify_buf_available();
+    void wait_buf_available();
+
+  protected:
+    // Configure the RDMA session
+    virtual Result configure(context::Context& ctx, const mcm_conn_param& request,
+                             const std::string& dev_port, libfabric_ctx *& dev_handle, Kind kind,
```
tszumski, line 63: Please set ```Kind kind``` inside RdmaRx/RdmaTx constructor, you dont have to pass it through ```Configure()``` method

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+
+    std::mutex cq_mutex;              // Mutex for completion queue synchronization
+    std::condition_variable_any cq_cv; // Condition variable for completion queue events
+    bool event_ready = false;         // Indicates if an event is ready in the CQ
+
+    void notify_cq_event();
+    void shutdown_rdma(context::Context& ctx);
+
+    //Helper functions
+    std::string kind_to_string(Kind kind);
```
ko80, line 119: You don't need it in the class, do you?

BTW we have the following in conn.cc:
```cpp
const char * kind2str(Kind kind, bool brief)
{
    switch (kind) {
    case Kind::undefined:   return brief ? "Undef" : "undefined";
    case Kind::transmitter: return brief ? "Tx" : "transmitter";
    case Kind::receiver:    return brief ? "Rx" : "receiver";
    default:                return str_unknown;
    }
}
```

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+#ifndef CQ_BATCH_SIZE
+#define CQ_BATCH_SIZE 64
+#endif
+#ifndef PAGE_SIZE
+#define PAGE_SIZE 4096
+#endif
+
+namespace mesh {
+
+namespace connection {
```
ko80, line 33: Change where applicable
```suggestion
namespace mesh::connection {
```

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,128 @@
...
+
+    //Helper functions
+    std::string kind_to_string(Kind kind);
+
+    std::atomic<bool> buf_available; // Indicates buffer availability in the queue
+};
+
+} // namespace connection
+
+} // namespace mesh
```
ko80, line 126: Change where applicable
```suggestion
} // namespace mesh::connection
```

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+    case Kind::undefined:
+        return "undefined";
+    case Kind::transmitter:
+        return "transmitter";
+    case Kind::receiver:
+        return "receiver";
+    default:
+        return "unknown";
+    }
+}
```
ko80, line 27: Replace with `kind2str()` from conn.cc.

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+{
+    buf_available.wait(false, std::memory_order_acquire);
+    buf_available = false;
+}
+
+// Add to queue
+Result Rdma::add_to_queue(void *element)
+{
+    if (!element) {
+        return Result::error_bad_argument;
```
ko80, line 54: Change where applicable.
```suggestion
        return set_result(Result::error_bad_argument);
```
This method automatically increments the error metric automatically on error results.

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// Consume from queue
+Result Rdma::consume_from_queue(context::Context& ctx, void **element)
+{
+    std::unique_lock<std::mutex> lock(queue_mutex);
+
+    if (ctx.cancelled()) {
+        *element = nullptr; // Ensure element remains nullptr
+        return Result::error_operation_cancelled;
```
ko80, line 73:
```suggestion
        return Result::error_context_cancelled;
```

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    // Calculate total memory size needed and align to page size
+    size_t aligned_trx_sz = ((trx_sz + PAGE_SIZE - 1) / PAGE_SIZE) * PAGE_SIZE;
+    size_t total_size = capacity * aligned_trx_sz;
+
+    // Allocate a single contiguous memory block
+    void *memory_block = std::aligned_alloc(PAGE_SIZE, total_size);
+    if (!memory_block) {
+        log::error("Failed to allocate a single memory block")("total_size", total_size);
```
ko80, line 120: Add details where this error happened (RDMA or whatever).

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+    if (trx_sz == 0 || trx_sz > MAX_BUFFER_SIZE) {
+        log::error("Invalid transfer size provided for buffer allocation")("trx_sz", trx_sz);
+        return Result::error_bad_argument;
+    }
+
+    std::lock_guard<std::mutex> lock(queue_mutex);
+
+    // Check if already initialized
+    if (!buffer_queue.empty()) {
+        log::error("Buffer queue already initialized");
```
ko80, line 109: Make the log message more descriptive. We have numerous components in Media Proxy, so the logs must bring more details.
```suggestion
        log::error("RDMA buffer queue already initialized");
```

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,388 @@
...
+ * @brief Shuts down the RDMA connection and cleans up resources.
+ *
+ * This function cancels the RDMA completion queue thread and the buffer processing thread,
+ * ensuring that they are properly joined. It also notifies buffer availability to unblock
+ * any waiting operations. If the RDMA connection was initialized, it cleans up the RDMA
+ * resources and sets the state to closed.
+ *
+ * @param ctx The context used for the shutdown operation.
+ */
+void Rdma::shutdown_rdma(context::Context& ctx)
```
tszumski, line 359: It looks like you can move content of ```Rdma::shutdown_rdma()```  into ```Rdma::on_shutdown()```

media-proxy/src/mesh/conn_rdma_tx.cc
```diff
@@ -0,0 +1,186 @@
+#include "conn_rdma_tx.h"
+#include <stdexcept>
+#include <queue>
+
+namespace mesh {
+
+namespace connection {
+
+RdmaTx::RdmaTx() : Rdma() { init_buf_available(); }
```
tszumski, line 9: You dont have to call ```Rdma()```

media-proxy/src/mesh/conn_rdma_rx.cc
```diff
@@ -0,0 +1,149 @@
+#include "conn_rdma_rx.h"
+#include <stdexcept>
+#include <queue>
+
+namespace mesh {
+
+namespace connection {
+
+RdmaRx::RdmaRx() : Rdma() { init_buf_available(); }
```
tszumski, line 9: You dont have to call Rdma()

tszumski: Also move call to ```init_buf_available()``` into ```on_establish()```

media-proxy/src/mesh/conn_rdma_rx.cc
```diff
@@ -0,0 +1,149 @@
...
+                    // Notify that a buffer is available
+                    notify_buf_available();
+                } else {
+                    log::error("Failed to add buffer back to the queue")("buffer_address", buf)
+                              (" ", kind_to_string(_kind));
+                }
+            }
+        } else if (ret == -EAGAIN) {
+            // No events to process, yield CPU briefly to avoid busy looping
+            std::this_thread::sleep_for(std::chrono::microseconds(CQ_RETRY_DELAY_US));
```
tszumski, line 138: Add new function inside concurrency.cc
```
void Sleep(context::Context& ctx, std::chrono::microseconds interval_us)
{
    std::mutex mx;
    std::unique_lock<std::mutex> lk(mx);
    std::condition_variable_any cv;
    cv.wait_for(lk, ctx.stop_token(), interval_us, [] { return false; });
}
```
And replace usage of ```this_thread::sleep``` with:

media-proxy/src/mesh/conn_rdma_tx.cc
```diff
@@ -0,0 +1,186 @@
...
+                    Result res = add_to_queue(buf);
+                    if (res != Result::success) {
+                        log::error("Failed to add buffer back to queue")("buffer_address", buf)
+                                  (" ", kind_to_string(_kind));
+                    }
+                }
+                break; // Exit retry loop as CQ events were successfully processed
+            } else if (ret == -EAGAIN) {
+                // No events, introduce short wait to avoid busy looping
+                std::this_thread::sleep_for(std::chrono::microseconds(RETRY_INTERVAL_US));
```
tszumski, line 83: Same here, replace sleep with ```void Sleep(context::Context& ctx, std::chrono::microseconds interval_us)```

media-proxy/src/mesh/conn_rdma_tx.cc
```diff
@@ -0,0 +1,186 @@
...
+                  ("trx_sz", trx_sz)(" ", kind_to_string(_kind));
+    }
+
+    std::memcpy(reg_buf, ptr, tmp_sent);
+
+    // Transmit the buffer through RDMA
+    int err = libfabric_ep_ops.ep_send_buf(ep_ctx, reg_buf, tmp_sent);
+    notify_buf_available();
+    sent = tmp_sent;
+    total_sent += tmp_sent;
```
tszumski, line 168: Is ```total_sent``` required ?

media-proxy/src/mesh/conn_rdma_rx.cc
ko80: Well, I don't think we need a delay of less than a millisecond to be cancellable.
@moleksy Keep it as it is for microsecond delays.

ko80: Monitoring if the context has been cancelled is needed only on operations blocking for 100ms or more. In other cases it is safe to use `this_thread::sleep_for`.

tszumski: Sound reasonable

media-proxy/include/mesh/conn_rdma.h
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/include/mesh/conn_rdma.h
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/include/mesh/conn_rdma.h
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma_tx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma_tx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/1e7f66891946f5a904408fb568ca119b7e04eaa0

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,359 @@
...
+ * Ensures proper cleanup and state updates on failure. Returns success if all steps
+ * (device initialization, endpoint setup, buffer queue creation, and endpoint configuration)
+ * complete successfully.
+ *
+ * @param ctx The operation context.
+ * @return Result::success on success, or an appropriate error result on failure.
+ */
+Result Rdma::on_establish(context::Context& ctx)
+{
+    if (init) {
```
Sakoram, line 187: Is it possible that:
- multiple threads will call on_establish() at the same time?
- one thread call on_establish() and the second calls shortly after on_shutdown()?
If yes, we need a better lock then checking `init`

I suspect that multiple threads calling on_establish() is very possible, as currently every connection to proxy creates its own thread

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,359 @@
...
+        return Result::error_already_initialized;
+    }
+
+    init_buf_available();
+
+    int ret;
+    Result res;
+
+    if (!mDevHandle) {
+        ret = libfabric_dev_ops.rdma_init(&mDevHandle);
```
Sakoram, line 199: I think, this code should not only allocate RDMA device with `rdma_init()`, but also deallocate it with rdma_deinit().
When? dunno, maybe when all RDMA sessions are closed, but I think it should be handled.

media-proxy/src/libfabric_ep.c
```diff
@@ -138,14 +139,28 @@ int ep_reg_mr(ep_ctx_t *ep_ctx, void *data_buf, size_t data_buf_size)
...
+    }
+
     int ret;
 
     do {
-        ret = fi_send(ep_ctx->ep, buf, buf_size, ep_ctx->data_desc, ep_ctx->dest_av_entry, NULL);
-        if (ret == -EAGAIN)
-            (void)fi_cq_read(ep_ctx->cq_ctx.cq, NULL, 0);
+         // Check if the stop flag is set
+        if (ep_ctx->stop_flag) {
```
Sakoram, line 153: As you added this flag my assumption is one thread sets it to let know an other thread that sending is unavailable. 
In this case volatile bool without any mutex seems insufficient to ensure that fi_send won't be called after `stop_flag` was set to true.

if this is done by single thread I think we have to allow to call on_receive only when the connection is in appropriate state. No flag would be needed then.

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,359 @@
...
+ * 
+ * @param ctx The operation context.
+ * @param request The connection parameters including addresses and RDMA arguments.
+ * @param dev_port The device port identifier.
+ * @param dev_handle Reference to the RDMA device handle.
+ * @param kind The type of connection (receiver or transmitter).
+ * @param dir The data transfer direction (RX or TX).
+ * @return Result::success on successful configuration, or an error result if arguments are invalid.
+ */
+Result Rdma::configure(context::Context& ctx, const mcm_conn_param& request,
```
Sakoram, line 153: is `dev_port` used in any way? if yes, why is it outside mcm_conn_param, if no, why does it exist?

media-proxy/src/mesh/conn_rdma.cc
ko80: No, there is only one decision maker in Media Proxy that calls `establish()` – BridgesManager. The issue won't happen by design.

media-proxy/src/mesh/conn_rdma_rx.cc
```diff
@@ -0,0 +1,149 @@
...
+        // Process all available buffers in the queue
+        while (!ctx.cancelled()) {
+            Result res = consume_from_queue(ctx, &buf);
+            if (res == Result::success && buf != nullptr) {
+                int err = libfabric_ep_ops.ep_recv_buf(ep_ctx, buf, trx_sz, buf);
+                if (err) {
+                    log::error("Failed to pass empty buffer to RDMA to receive into")
+                              ("buffer_address", buf)("error", fi_strerror(-err))
+                              (" ",kind2str(_kind));
+                    add_to_queue(buf);
```
Sakoram, line 72: add_to_queue() returns a return value, check it.

media-proxy/src/mesh/conn_rdma_tx.cc
```diff
@@ -0,0 +1,180 @@
...
+    int err = libfabric_ep_ops.ep_send_buf(ep_ctx, reg_buf, tmp_sent);
+    notify_buf_available();
+    sent = tmp_sent;
+
+    if (err) {
+        log::error("Failed to send buffer through RDMA")("error", fi_strerror(-err))
+                  (" ", kind2str(_kind));
+
+        // Add the buffer back to the queue in case of failure
+        add_to_queue(reg_buf);
```
Sakoram, line 171: add_to_queue() returns a return value, check it.

media-proxy/src/session-rdma-rx.cc
```diff
@@ -92,7 +92,6 @@ RxRdmaSession::RxRdmaSession(libfabric_ctx *dev_handle, const mcm_conn_param &re
     ep_cfg.rdma_ctx = dev_handle;
     memcpy(&ep_cfg.remote_addr, &request.remote_addr, sizeof(request.remote_addr));
     memcpy(&ep_cfg.local_addr, &request.local_addr, sizeof(request.local_addr));
-    ep_cfg.dir = direction::RX;
```
Sakoram, line 95: are you sure that this does not break the current implementation?

media-proxy/src/session-rdma-tx.cc
```diff
@@ -49,7 +49,6 @@ TxRdmaSession::TxRdmaSession(libfabric_ctx *dev_handle, const mcm_conn_param &re
     ep_cfg.rdma_ctx = dev_handle;
     memcpy(&ep_cfg.remote_addr, &request.remote_addr, sizeof(request.remote_addr));
     memcpy(&ep_cfg.local_addr, &request.local_addr, sizeof(request.local_addr));
-    ep_cfg.dir = direction::TX;
```
Sakoram, line 52: are you sure that this does not break the current implementation?

media-proxy/src/mesh/conn_rdma.cc
moleksy: For now I was calling ep_destroy upon shutdown, the initialized device was ready to be called upon again, I can add 

    if (mDevHandle) {
        libfabric_dev_ops.rdma_deinit(&mDevHandle);
    }
    
  In cleanup_resources if you think it's a better approach.

media-proxy/src/libfabric_ep.c
moleksy: Flag is needed for a gracefull shutdown, if device was replaying with -EAGAIN constantly (when on the other end nothing was received), the thread was working endlessly.

media-proxy/src/mesh/conn_rdma.cc
moleksy: It's in place to allow other type of device identification than pure ip address, this proves to be problematic in some scenarios, verbs sometimes tends to treat passed ip address as suggestion not requirement for choosing a device

media-proxy/src/mesh/conn_rdma_rx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/737483c86a4bf88fdd86b9e223e58acf7f3b7d6d

media-proxy/src/mesh/conn_rdma_tx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/737483c86a4bf88fdd86b9e223e58acf7f3b7d6d

media-proxy/src/session-rdma-tx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/commit/737483c86a4bf88fdd86b9e223e58acf7f3b7d6d

media-proxy/src/mesh/conn_rdma.cc
Sakoram: It is great if you can use PCI port to select the device. My point is that I can't see that happening here. `dev_port` is taken as parameter to `configure()`, but it is never used.

media-proxy/src/mesh/conn_rdma.cc
Sakoram: No, I don't think adding `rdma_deinit` in `cleanup_resources()` is a good choice. 
We can have multiple concurrent RDMA connections, one connection should not deallocate whole libfabric context... unless it is sure that no other connection exists.

I guess the deallocation should be done by some other, higher abstraction that manages all connections. This will be done probably only at shutdown.

But even though it is likely that the higher level will call it, I would encapsulate `rdma_deinit` into some function here so as not to call the libfabric implementation directly and to maintain separation.

moleksy: Perhaps introducing a counter at rdma base class would be ok?
I'd only call on rdma_deinit if the active_connection counter would be 0 (after shutting down the last/or the only one).

moleksy: something like `static std::atomic<int> active_connections;`

Sakoram: It would work. You can do it if you feel it is good approach. I only have 2 questions:
- Do we want to always deallocate libfabric resources when there is 0 connections? If for some reason this is done frequently, some delays will be introduced.
- Do we want connection class to be responsible for this resource? This may be the case, but an MCM agent, for example, would be more aware of when it should actually be initialized and destroyed, but I think the agent would not be necessary and always present. 

idk. I do not insist on any particular solution. I only know that if we allocate something we should clean it somewhere.

ko80: The Agent always exists and accessible for Media Proxies. If the Agent is unavailable, the SDK won't create connections.

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,359 @@
...
+        log::error("RDMA configuring failed")("state", "closed");
+        if (ep_ctx) {
+            libfabric_ep_ops.ep_destroy(&ep_ctx);
+        }
+        set_state(ctx, State::closed);
+        return res;
+    }
+
+    init = true;
+    res = start_threads(ctx);
```
Sakoram, line 243: res is never checked

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,359 @@
...
+    // Cancel `rdma_cq_thread` context
+    rdma_cq_thread_ctx.cancel();
+    // Cancel `process_buffers_thread` context
+    process_buffers_thread_ctx.cancel();
+    // Ensure buffer-related threads are unblocked
+    notify_buf_available();
+
+    // Check if `rdma_cq_thread` has stopped
+    if (handle_rdma_cq_thread.joinable()) {
+        handle_rdma_cq_thread.join();
```
Sakoram, line 271: join() can throw an exception. low chances for that, but I think we wanted to catch all exceptions and return errors by return value.

Sakoram: I suggest to mark all methods as `noexcept`

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/13dda25882a90d3e6530502cc56dd0439b96ff95

moleksy: Deinitialization should happen sparsely, especially if there would be many connections on the same device. Possibly, measurable impact would be only in stress testing when there would be a lot single connections created and destroyed one by one as fast as possible.

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/c49f5eba0ccf59ae374fbfcf4cdfd4888363324d

media-proxy/src/mesh/conn_rdma.cc
moleksy: Can be added in future PRs

moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/390a65332407a21263d7269b722310576780ad64

media-proxy/src/mesh/conn_rdma.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/390a65332407a21263d7269b722310576780ad64

media-proxy/include/mesh/conn_rdma.h
```diff
@@ -0,0 +1,117 @@
...
+
+    // Queue for managing buffers
+    std::queue<void*> buffer_queue; // Queue holding available buffers
+    std::mutex queue_mutex;         // Mutex for buffer queue synchronization
+    std::condition_variable_any queue_cv; // Condition variable for buffer availability
+
+    // // RDMA thread logic
+    virtual Result start_threads(context::Context& ctx) = 0;
+
+    Result init_queue_with_elements(size_t capacity, size_t trx_sz);
```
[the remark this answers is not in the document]
Sakoram: I'm always worry that ‘in future PRs’ is a polite version of saying ‘never’ ^^

media-proxy/src/mesh/conn_rdma.cc
```diff
@@ -0,0 +1,399 @@
...
+
+std::atomic<int> Rdma::active_connections(0);
+
+Rdma::Rdma() : ep_ctx(nullptr), init(false), trx_sz(0), mDevHandle(nullptr) {
+    std::memset(&ep_cfg, 0, sizeof(ep_cfg));
+    ++active_connections;
+}
+
+Rdma::~Rdma() {
+    on_shutdown(context::Background());
```
Sakoram, line 14: on_shutdown return a return value. It is not checked here and in other places.

Don't we have any warnings enabled that would inform us about things like that? something like `-Wunused-result`. Could you possibly enable those warnings? In my opinion we should enable all warnings and threat them as errors.

media-proxy/include/mesh/conn_rdma.h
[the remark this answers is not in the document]
ko80: I strongly suggest to avoid pushing the developer that way. We all care about the code and the project, but most importantly we value our own time invested in the development. This RDMA effort needs to be finalized as soon as possible to enable integration with other subsystems in Media Proxy. Future PRs will follow quite shortly when we have the very first solution of Multipoint Groups working in basic scenarios.

@Sakoram Could you complete your review focusing on critical issues only?

media-proxy/src/mesh/conn_rdma_rx.cc
```diff
@@ -0,0 +1,154 @@
...
+                    log::error("Failed to transmit buffer")("buffer_address", buf)("size", trx_sz)
+                              (" ", kind2str(_kind));
+                    continue;
+                }
+
+                // Add the buffer back to the queue
+                res = add_to_queue(buf);
+                if (res == Result::success) {
+                    // Notify that a buffer is available
+                    notify_buf_available();
```
Sakoram, line 136: notify_buf_available() is already called inside add_to_queue()

ko80:
I suppose the PR is ready to be merged. Let's complete the review and approve.

tszumski (approved):

ko80 (dismissed):

media-proxy/src/mesh/conn_rdma_rx.cc
ko80: @moleksy Please include this in future refactoring.

media-proxy/src/mesh/conn_rdma.cc
ko80: @moleksy Please include this in future refactoring.

Sakoram (approved):

tszumski (approved):

CMakeLists.txt
```diff
@@ -35,13 +35,23 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 add_subdirectory(${SDK_DIR})
 add_subdirectory(${MP_DIR})
 
+
+# Fetch google test dependencies
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/35d0c365609296fa4730d62057c487e3cfa030ff.zip
```
ko80, line 43: Can you add a comment here why this exact URL is used here? And why we can't install the dependency in the installation phase, not the build phase?

media-proxy/include/mesh/conn.h
```diff
@@ -67,8 +67,14 @@ enum class Result {
     error_bad_argument,
     error_out_of_memory,
     error_general_failure,
+    error_already_initialized,
+    error_initialization_failed,
+    error_memory_registration_failed,
+    error_thread_creation_failed,
+    error_no_buffer,
+    error_timeout,
```
ko80, line 75: This should go after `error_context_cancelled`

media-proxy/src/mesh/conn.cc
```diff
@@ -311,6 +311,13 @@ const char * result2str(Result res)
     case Result::error_bad_argument:     return "bad argument";
     case Result::error_out_of_memory:    return "out of memory";
     case Result::error_general_failure:  return "general failure";
+    case Result::error_context_cancelled:          return "context cancelled";
+    case Result::error_already_initialized:        return "already initialized";
+    case Result::error_initialization_failed:      return "initialization failed";
+    case Result::error_memory_registration_failed: return "memory registration failed";
+    case Result::error_thread_creation_failed:     return "thread creation failed";
+    case Result::error_no_buffer:        return "no buffer";
```
ko80, line 319: These all are subject to remove (rework) in the future refactoring. Reason: these bring details that are related to the particular derived class, not the base class. No need to change anything for now.

sdk/include/mcm_dp.h
```diff
@@ -207,6 +207,7 @@ typedef struct {
 /* rdma format */
 typedef struct {
     size_t transfer_size;
+    int queue_size;
```
ko80, line 210: Does this change affect the SDK API declared in `mesh_dp.h`?

CMakeLists.txt
moleksy: Change has been proposed by CI/CD team.

media-proxy/src/mesh/conn_rdma_rx.cc
```diff
@@ -0,0 +1,154 @@
...
+    } catch (const std::system_error& e) {
+        log::fatal("Failed to start thread")("error", e.what());
+        return Result::error_thread_creation_failed;
+    }
+
+    try {
+        handle_rdma_cq_thread =
+            std::jthread([this]() { this->rdma_cq_thread(this->rdma_cq_thread_ctx); });
+    } catch (const std::system_error& e) {
+        log::fatal("Failed to start thread")("error", e.what());
```
ko80, line 41: The use of `log::fatal` here confuses me because I understand that it exists the process right away.

ko80: Can you replace `log::fatal` with `log::error` everywhere?

sdk/include/mcm_dp.h
moleksy: mcm_rdma_args is not used inside mesh_dp.h

media-proxy/include/mesh/conn.h
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/794708adf96f927afbd18cdc93def4203541ba83

media-proxy/src/mesh/conn_rdma_rx.cc
moleksy: fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/286/commits/794708adf96f927afbd18cdc93def4203541ba83

media-proxy/src/mesh/conn_rdma_tx.cc
```diff
@@ -28,7 +28,7 @@ Result RdmaTx::start_threads(context::Context& ctx)
         handle_rdma_cq_thread =
             std::jthread([this]() { this->rdma_cq_thread(this->rdma_cq_thread_ctx); });
     } catch (const std::system_error& e) {
-        log::fatal("Failed to start threads")("error", e.what())(" ", kind2str(_kind));
+        log::error("Failed to start threads")("error", e.what())(" ", kind2str(_kind));
```
ko80, line 31: Still the parameter is noname

ko80 (approved):
Looks good to me

tszumski (approved):

PR #292, opened 2024-12-17
Update ST2110 device parameters to reduce memory/time overhead

tszumski:
Based on @DawidWesierski4 feedback we don't need more than 17 queue count, in result memory overhead and init time is reduced significantly

### DIFF
--- a/media-proxy/src/mesh/st2110.cc
+++ b/media-proxy/src/mesh/st2110.cc
@@ -125,14 +125,12 @@ void get_mtl_dev_params(mtl_init_params& st_param, const std::string& dev_port,
     st_param.log_level = log_level;
     st_param.priv = NULL;
     st_param.ptp_get_time_fn = NULL;
-    // Native af_xdp have only 62 queues available
-    if (st_param.pmd[MTL_PORT_P] == MTL_PMD_NATIVE_AF_XDP) {
-        st_param.rx_queues_cnt[MTL_PORT_P] = 62;
-        st_param.tx_queues_cnt[MTL_PORT_P] = 62;
-    } else {
-        st_param.rx_queues_cnt[MTL_PORT_P] = 128;
-        st_param.tx_queues_cnt[MTL_PORT_P] = 128;
-    }
+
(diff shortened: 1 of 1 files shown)


DawidWesierski4:
Yes mtl backend supports up to 16 schedulers 
without SHARED QUEUES flag every scheduler "gets" one queue 

without support for more schedulers assigning 128 queues is just a waste of resources 

when we will introduce support for more then 16 schedulers the conversation about increasing this number can come back 


every scheduler takes 
1 core from dpdk backend 
1 queue to work with data

DawidWesierski4 (commented):
LGTM 
But i would use a define for that purpose

media-proxy/src/mesh/st2110.cc
```diff
@@ -125,14 +125,8 @@ void get_mtl_dev_params(mtl_init_params& st_param, const std::string& dev_port,
...
-    // Native af_xdp have only 62 queues available
-    if (st_param.pmd[MTL_PORT_P] == MTL_PMD_NATIVE_AF_XDP) {
-        st_param.rx_queues_cnt[MTL_PORT_P] = 62;
-        st_param.tx_queues_cnt[MTL_PORT_P] = 62;
-    } else {
-        st_param.rx_queues_cnt[MTL_PORT_P] = 128;
-        st_param.tx_queues_cnt[MTL_PORT_P] = 128;
-    }
+    st_param.rx_queues_cnt[MTL_PORT_P] = 17;
+    st_param.tx_queues_cnt[MTL_PORT_P] = 17;
```
ko80, line 129: Please leave a visible comment before these lines to explain why these seventeens are hardcoded here.

tszumski: Fixed in 464e73822b867e18a244efc9a09613f678f0a633

DawidWesierski4 (approved):
Lgtm

ko80 (approved):

moleksy (approved):
LGTM

PR #298, opened 2024-12-20
Add proposal for json video format mappings

tszumski:
Add table that map json string to ST2110/FFmpeg format enums

### DIFF
--- a/docs/sdk-json-proposal/SDK_API_WORKFLOW.md
+++ b/docs/sdk-json-proposal/SDK_API_WORKFLOW.md
@@ -62,6 +62,14 @@
   }
 }
 ```
+| BCS(json string) | MTL-transport_format(enum st20_fmt) | MTL {in/out}put_format(enum st_frame_fmt) | FFMPEG (enum PixelFormat) | RFC 4715 compliant | Note |
+|------------------|-------------------------------------|-------------------------------------------|---------------------------|--------------------|------|
+| y210le | ST20_FMT_YUV_422_10BIT | ST_FRAME_FMT_Y210 |AV_PIX_FMT_Y210LE | ?? | -- |
+| yuv422p10le | ST20_FMT_YUV_422_PLANAR10LE | ST_FRAME_FMT_YUV422PLANAR10LE | PIX_FMT_YUV422P10LE | N | -- |
+| yuv422p10le_rfc4715 | ST20_FMT_YUV_422_10BIT | ST_FRAME_FMT_YUV422RFC4175PG2BE10 | N/A | Y | -- |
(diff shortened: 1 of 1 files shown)


ko80:
Please add another column showing the corresponding string name constant used in FFmpeg.

Reference: `libavutil/pixdesc.c`
```c
static const AVPixFmtDescriptor av_pix_fmt_descriptors[AV_PIX_FMT_NB]
```

Mionsz (commented):
So my question is if this shouldn't be moved downstream to MTL and introduce it in one place? Also would be great if BCS and FFMPEG would use the FFMPEG naming convention and not introduce new overhead.

https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/pixdesc.c

so for `yuv422p10le` example:

image

This would also allow us to have one source of truth - MTL and unified naming convention - I believe that so many ears of development for FFMPEG allowed them to evolve to the best possible set of ENUM <--> STRING color definitions.

tszumski:
Please keep in mind that yuv422p10le is not natively supported by ST2110 standard

Mionsz:
So there are ready methods for evaluating pacing/packing/
image

image


> Please keep in mind that yuv422p10le is not natively supported by ST2110 standard

@tszumski this was just pasted as an example. We will require dozens of formats - and `yuv422p10le` is not the important one in the end.

docs/sdk-json-proposal/SDK_API_WORKFLOW.md
```diff
@@ -62,6 +62,14 @@
   }
 }
 ```
+| BCS(json string) | MTL transport_format(enum st20_fmt) | MTL {in/out}put_format(enum st_frame_fmt) | FFmpeg (enum PixelFormat) |FFmpeg (string name) | RFC 4715 compliant | Note |
```
MateuszGrabuszynski, line 65: [RFC 4715](https://datatracker.ietf.org/doc/rfc4715/) has a title *The Integrated Services Digital Network (ISDN) Subaddress Encoding Type for tel URI*. Is it the right document?

tszumski:
> So there are ready methods for evaluating pacing/packing/ image
> 
> image
> 
> > Please keep in mind that yuv422p10le is not natively supported by ST2110 standard
> 
> @tszumski this was just pasted as an example. We will require dozens of formats - and `yuv422p10le` is not the important one in the end.

Those methods are not exposed in MTL API, also we use different C++ json parser, so they are not too helpful

docs/sdk-json-proposal/SDK_API_WORKFLOW.md
tszumski: Oh, sorry my bad, it should be 4175 _RTP Payload Format for Uncompressed Video_

MateuszGrabuszynski: Thank you :)

PR #300, opened 2024-12-31
Implement JSON configuration in SDK

ko80:
* Add support for JSON configuration in SDK.
* Upgrade FFmpeg video and audio plugins to support JSON config.
* Propagate SDK connection config to Agent, multipoint groups, and bridges.
* Make API ports configurable from command line.
* Fix CI/CD.

PR #301, opened 2025-01-07
Removal of unnecessary requirements.txt and requirements.in files in media-proxy folder

MateuszGrabuszynski:
Removal of unnecessary `requirements.txt` and `requirements.in` files in `media-proxy` folder. Those files raise CVE issues for outdated and deprecated packages.

1. The file’s header (comment at the very beginning) mentions it is automatically generated using a command. It seems someone may have ran the command in the wrong directory. 
2. It seems like there’s no Python code inside that folder, but only C/C++, thus the pip packages list provided does not seem to have any meaning at all.
3. The file was added in December 2023, and only the versions of the packages were bumped from then, mostly based on Dependabot suggestions.


ko80:
This is to be merged AFTER https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/300

soopel (approved):
Approved

ko80 (approved):

PR #306, opened 2025-01-13
Extend MCM API implementation in Media Proxy and SDK

tszumski:
1. Extend MCM API implementation in Media Proxy and SDK with new ST2110 Transport Pixel Format field
2. Update MCM pixelFormat based on architect feedback


sdk/include/mesh_dp.h
```diff
@@ -124,6 +124,14 @@ typedef struct MeshConfig_ST2110 {
...
 #define MESH_CONN_TRANSPORT_ST2110_30 2 ///< SMPTE ST2110-30 Audio transport
 
+    /**
+     * SMPTE ST2110-20 rfc4175 compliant transport format.
+     * Required only for ST2110-20 transport.
+     * Any value of the MESH_CONN_ST2110_20_TRANSPORT_FMT_* constants.
+     */
+    int transport_format;
+#define MESH_CONN_ST2110_20_TRANSPORT_FMT_YUV422_10BIT 0 ///< YUV 4:2:2 10-bit, "yuv422p10rfc4175"
+
```
ko80, line 134: What is the benefit of specifying the transport pixel format if there is the only supported option? I don't get it.

media-proxy/src/mesh/st2110.cc
```diff
@@ -10,20 +10,26 @@ namespace mesh::connection {
...
-    case MESH_VIDEO_PIXEL_FORMAT_RGB8:
-        st_fmt = ST_FRAME_FMT_RGB8;
+    default:
+        return -1; // Error: unknown format
+    }
+
+    return 0; // Success
+}
+
+int mesh_transport_video_format_to_st20_fmt(int transport_format, st20_fmt& st20_format) {
```
ko80, line 29: If there is the only transport pixel format option, why not to set `st20_format = ST20_FMT_YUV_422_10BIT` in all scenarios?

ko80:
Please notice you need to update the video frame calculation in `calc_payload_size()` according to the introduced pixel format changes.

sdk/include/mesh_dp.h
tszumski: Based on the architect's feedback, for the Q1 release, we want to have one working pixel format that can be extended in the future. By specifying that option in the API (JSON), we also avoid yet another API mismatch between releases.

tszumski:
> Are you going to add parsing of `transportPixelFormat` in `parse_json()`?

Fixed in 9b32813b51f2e44ae1eac04a02b89be153f3e78f

tszumski:
> Please notice you need to update the video frame calculation in `calc_payload_size()` according to the introduced pixel format changes.

Fixed in 9b32813b51f2e44ae1eac04a02b89be153f3e78f

media-proxy/src/mesh/st2110.cc
tszumski: While setting st20_format = ST20_FMT_YUV_422_10BIT in all scenarios is a valid option, I prefer to keep parameter parsing and passing flexible. This way, we can easily accommodate any future changes or requests from the architects without needing to make significant adjustments.

ko80: I suggest adding a detailed comment here in the code with strong arguments why this is done this way.

sdk/src/mesh_conn.cc
```diff
@@ -340,12 +340,36 @@ int ConnectionJsonConfig::calc_audio_buf_size()
...
+    } else if (payload.video.pixel_format == MESH_VIDEO_PIXEL_FORMAT_YUV422RFC4175BE10) {
+        if (pixels % 2) {
+            log::error("Invalid width %u height %u for yuv422rfc4175be10 fmt, not multiple of 2",
+                       payload.video.width, payload.video.height);
+            return -MESH_ERR_CONN_CONFIG_INVAL;
+        }
+        calculated_payload_size = pixels * 5 / 2;
+    } else
+        return -MESH_ERR_CONN_CONFIG_INVAL;
+    return 0;
```
ko80, line 365: Could you rework this to use `switch/case`?

sdk/src/mesh_conn.cc
```diff
@@ -92,6 +92,10 @@ int ConnectionJsonConfig::parse_json(const char *str)
 
             conn.st2110.pacing = st2110.value("pacing", "");
             conn.st2110.payload_type = st2110.value("payloadType", 112);
+            if (conn.st2110.transport == MESH_CONN_TRANSPORT_ST2110_20) {
```
ko80, line 95: The `parse_json()` function is covered by unit tests. Please amend the unit tests to cover the added functionality.

ko80:
Can you check if the traffic goes?


I'm getting an error reported by `ST2110_20Tx->configure()`
image

Command line:
```bash
sudo ffmpeg -stream_loop -1 -re -video_size 640x360 -pixel_format yuv422p10le -i netflix_crosswalk_640x360_60fps_10bit_le.yuv -f mcm -conn_type st2110 -transport st2110-20 -ip_addr 192.168.96.11 -port 9002 -frame_rate 60 -video_size 640x360 -pixel_format yuv422p10le -
```

sdk/src/mesh_conn.cc
tszumski: Fixed in ce2adbe5d9fc31daa6158e501ad49cac1b0582d4

sdk/src/mesh_conn.cc
tszumski: Fixed in e2d027c50e0c3acf589153f1a5b40878eb7d1c83

media-proxy/src/mesh/st2110.cc
tszumski: Fixed in 7ff9910aea5decc2d4da77ab9e18f843031802ae

ko80 (approved):
Looks good to me

moleksy (approved):
Looks Good

PR #322, opened 2025-01-29
Val: Switching mock app to Łappka in pytest

MateuszGrabuszynski:
Switches mock SDK testing application to Łappka (application using real Media Communications Mesh SDK API) with the most basic tests for now.

Fix: mcm.c switch from unsigned char (0-255) to unsigned int as a frame counter
Fix: multipoint-group -> multipointGroup in JSON generator
Fix: localhost address and default port ensure stable connection for a single media_proxy test
Add: mesh-agent starts before media_proxy
Add: fixed sleep-based timing for application runtime (ensures stability)
Add: more logging
Add: integrity tests from MTL (from @staszczuk)
Add: additional file removal function for pathlib-based paths (with backwards compatibility)

**To-do:**
- [x] `calculate_yuv_frame_size(...)` must use proper values
- [x] `pixelFormat` is based on ambiguous `media_files.py` (lack of data whether `YUV422PLANAR10LE` or `YUV422RFC4175PG2BE10`)
- [x] logging the output of each executed component (not parsing it in this PR, yet)
- [x] (optional) cleanup of `integrity.py` - could utilize a single way (function) of checking the chunk hashes for all types of media


tests/validation/Engine/engine_mcm.py
```diff
@@ -5,6 +5,9 @@
 import logging
 import signal
 import subprocess
+import time
+
+from pathlib import Path
```
MateuszGrabuszynski, line 10:
```suggestion
```

tests/tools/TestApp/src/mcm.c
```diff
@@ -24,7 +24,7 @@ int mcm_send_video_frames(MeshConnection *connection, const char *filename) {
         return err;
     }
 
-    unsigned char frame_num = 0;
```
zLukas, line 27: why this change ?
do you except to have more than 255 frames ?

tests/validation/Engine/connection_json.py
```diff
@@ -14,25 +14,30 @@ class ConnectionJson:
...
         maxMetadataSize=8192,
         connection=Connection(),
         payload=Payload(),
     ):
         self.bufferQueueCapacity = bufferQueueCapacity
-        self.maxPayloadSize = maxPayloadSize
+        if maxPayloadSize > 0:
+            self.maxPayloadSize = maxPayloadSize
+        else:
+            self.maxPayloadSize = None
```
zLukas, line 26: None or 0 ?  __init__ sets to 0 not None

tests/validation/Engine/engine_mcm.py
```diff
@@ -63,14 +75,44 @@ def remove_sent_file(file_path: str, app_path: str) -> None:
...
+        connection_cfg_file = Path(app_path.resolve(), "connection.json")
+        output_file_path = Path(app_path, Path(file_path).name + "_MCMoutput.yuv").resolve()
+        rx = run_rx_app(
+            client_cfg_file=client_cfg_file,
+            connection_cfg_file=connection_cfg_file,
+            path_to_output_file=str(output_file_path),
+            cwd=app_path,
+            timeout=timeout
+            )
+        time.sleep(2)
```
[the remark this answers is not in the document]
MateuszGrabuszynski: Added comments for that.

tests/validation/Engine/engine_mcm.py
```diff
@@ -63,14 +75,44 @@ def remove_sent_file(file_path: str, app_path: str) -> None:
...
 
-def run_rx_tx_with_file(file_path: str, build: str):
+def remove_sent_file(full_path: Path) -> None:
+    try:
+        full_path.unlink()
+        logging.debug(f"Removed: {full_path}")
+    # except makes the test pass if there's no file to remove
+    except (FileNotFoundError, NotADirectoryError):
+        logging.debug(f"Cannot remove. File does not exist: {full_path}")
+
```
zLukas, line 85: finally block ?

tests/validation/Engine/connection_json.py
```diff
@@ -14,25 +14,30 @@ class ConnectionJson:
...
         bufferQueueCapacity=16,
-        maxPayloadSize=2097152,
+        maxPayloadSize=0,
         maxMetadataSize=8192,
         connection=Connection(),
         payload=Payload(),
     ):
         self.bufferQueueCapacity = bufferQueueCapacity
-        self.maxPayloadSize = maxPayloadSize
+        if maxPayloadSize > 0:
```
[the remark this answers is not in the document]
MateuszGrabuszynski: I am not sure if I understand well. This value is checked later when creating a dictionary.

tests/validation/Engine/engine_mcm.py
```diff
@@ -63,14 +75,44 @@ def remove_sent_file(file_path: str, app_path: str) -> None:
...
+            path_to_input_file=file_path,
+            cwd=app_path
+            )
+        time.sleep(5)
         handle_tx_failure(tx)
-    finally:
         stop_rx_app(rx)
-        # TODO: Add checks for transmission errors
-        remove_sent_file(file_path, app_path)
+    finally:
```
zLukas, line 111: no `except` block

tests/validation/Engine/client_json.py
```diff
@@ -11,7 +11,7 @@ class ClientJson:
     def __init__(
```
zLukas, line 11: unify type hinting, sometimes it's used sometimes not

tests/tools/TestApp/src/mcm.c
MateuszGrabuszynski: Yes, we definitely do want to transmit more than 255 frames. Also, even if we wouldn't, why use character type for integer values?

tests/validation/Engine/connection_json.py
MateuszGrabuszynski: Just to back my reasoning:
`0` -> put the value into JSON as 0
`None` -> do not put this value into JSON at all
Perhaps, the value is wrongly checked later on, and that's what you mean?

tests/validation/Engine/engine_mcm.py
MateuszGrabuszynski: `finally` block is optional. We do not want to do anything in this function regardless of the `try` part.

tests/validation/Engine/engine_mcm.py
MateuszGrabuszynski: Means: do not handle any issues different way than default.

tests/validation/Engine/connection_json.py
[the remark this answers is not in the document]
MateuszGrabuszynski: Ok, I re-thought what I wrote. This was completely wrong. I changed it.

tests/validation/Engine/connection_json.py
MateuszGrabuszynski: Ok, I re-thought what I wrote. This was completely wrong. I changed it.

tests/validation/Engine/client_json.py
MateuszGrabuszynski: I think it should be ok now ;)

tests/validation/Engine/engine_mcm.py
MateuszGrabuszynski: After rethinking that, it absolutely makes sense. It just tries to calculate the frame integrity and deletes the file, even if anything was wrong during the transmission.

zLukas: finally will be run regardless try status, that's why it's called finally :P
but I still insist to have except block for consistency, or for logging purposes:
```
except:
   log("this error occurs")
  finally
```
if run_tx_app() fails, it will never execute handle_tx_failure() and stop_rx_app()
should'tn be better to move handle_tx_failure() and stop_rx_app() to finally block ?

tests/validation/Engine/connection_json.py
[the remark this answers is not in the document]
MateuszGrabuszynski: After a quick sync with Andrzej, it was changed and tested today (for me it works as intended). Hopefully that's enough to resolve the conversation. :)

tests/validation/Engine/connection_json.py
MateuszGrabuszynski: It was changed again today and tested thoroughly. Hopefully now it meets the expectations. Sorry for the confusion.

tests/validation/Engine/engine_mcm.py
MateuszGrabuszynski: I understand your concern, but to close this PR - please, let's agree on not adding the exception handling (as it would be just printing the error again). I will try to figure out a better way in the next one.

zLukas (approved):

andremiszcz (approved):

andremiszcz (approved):

KarolinaPomian (approved):
lgtm
tests PASS

PR #329, opened 2025-02-07
Remove iavf driver from installation script

tszumski:
The IAVF driver is installed by default on the OS. Updating it may cause unexpected behavior. The driver is required by the Media Transport Library, which does not mention about installing/updating it


sdk/build.sh
```diff
@@ -27,4 +27,4 @@ if [[ $# -ne 0 ]]; then
 fi
 as_root make -C "${MCM_SDK_DIR}" install
 
-ln -s "${MCM_SDK_DIR}" "${SCRIPT_DIR}/build"
+ln -sf "${MCM_SDK_DIR}" "${SCRIPT_DIR}/build"
```
moleksy, line 30: Did we have problems without forcing it?

tszumski: Yes, there is error:
```ln: failed to create symbolic link```

moleksy (approved):
LGTM

ko80 (approved):

PR #315, opened 2025-01-20
FIX: Fixes to the installation scripts

Mionsz:
FIX: Fixes to the installation scripts:
- setup_ice_rdma .sh - now install most of the important things without errors on clean machine
- setup_rdma_env.sh - minor fixes to the script - more issues still needs to be addressed.

PR #331, opened 2025-02-11
Add missing include to fix gcc-13.3.0 compilation

tszumski:
Add missing include that cause compilation error when using gcc-13.3.0
```
sdk/include/mesh_client.h:21:10: error: ‘string’ in namespace ‘std’ does not name a type
   21 |     std::string api_version
```


moleksy (approved):
Looks good to me

tszumski:
Closing because I did not notice that this fix is already included in #315

PR #336, opened 2025-02-18
Added: Docker images build for MtlManager and mesh-agent

Mionsz:
Added: Docker images build for MtlManager and mesh-agent as an extension of standard media-proxy build workflow


ko80 (approved):
I hope this doesn't crash CI/CD

media-proxy/Dockerfile
```diff
@@ -162,8 +164,8 @@ RUN apt-get update --fix-missing && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
     groupadd -g 2110 vfio && \
-    groupadd -g 1001 imtl && \
-    useradd -m -s /bin/bash -G vfio,imtl -u 1002 mcm && \
+    groupadd -g 2001 mtl && \
+    useradd -m -s /bin/bash -G vfio,mtl -u 2002 mcm && \
```
tszumski, line 168: What is the purpose of changing 1001 to 2002 ?

tszumski (approved):

media-proxy/Dockerfile
Mionsz: As having 2 accounts in the system would fail install for 1001 (second UID). That is why I assumed it would be more safe to use higher number - 2001.

PR #339, opened 2025-02-24
Add: Workflow for create validation-tests.yml

Mionsz:
Add: Workflow for create validation-tests.yml

### DIFF
--- a/.github/workflows/validation-tests.yml
+++ b/.github/workflows/validation-tests.yml
@@ -1,0 +1,231 @@
+name: validation-bare-metal
+
+on:
+  # allow manually trigger
+  workflow_dispatch:
+    inputs:
+      branch-to-checkout:
+        type: string
+        default: 'main'
+        required: false
+        description: 'Branch name to use'
+      validation-iface-binding:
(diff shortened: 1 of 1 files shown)


.github/workflows/validation-tests.yml
```diff
@@ -0,0 +1,231 @@
+name: validation-bare-metal
+
+on:
+  # allow manually trigger
+  workflow_dispatch:
+    inputs:
+      branch-to-checkout:
```
moleksy, line 7: I think it's not used in the code, so no matter what the input is here, it won't change the flow.

Mionsz: Also worth mentioning is that this action runs on self-hosted runners only: `runs-on: [Linux, self-hosted, DPDK]` as it require E810 NIC cards and full access to a Ubuntu based machine - this is achieved using Humbold and SED networking of course :)

moleksy (approved):
LGTM

MateuszGrabuszynski (approved):
lgtm

PR #341, opened 2025-02-27
Deploy mesh-agent

Magda-Pytel:
- Added another target for building in Dockerfile for mesh agent
- Updated readme with instructions how to build image of mesh agent

### DIFF
--- a/media-proxy/Dockerfile
+++ b/media-proxy/Dockerfile
@@ -187,8 +187,8 @@ HEALTHCHECK --interval=30s --timeout=5s CMD ps aux | grep "media_proxy" || exit 
 FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME} as mesh-agent
 
 LABEL maintainer="milosz.linkiewicz@intel.com"
-LABEL org.opencontainers.image.title="Media Communications Mesh Media Proxy Agent"
-LABEL org.opencontainers.image.description="Media Communications Mesh Media Proxy Agent application. Ubuntu 22.04 Docker Container Release Image"
+LABEL org.opencontainers.image.title="Media Communications Mesh Mesh Agent"
+LABEL org.opencontainers.image.description="Media Communications Mesh Mesh Agent application. Ubuntu 22.04 Docker Container Release Image"
 LABEL org.opencontainers.image.version="1.0.0"
 LABEL org.opencontainers.image.vendor="Intel Corporation"
 LABEL org.opencontainers.image.licenses="BSD 3-Clause License"
--- a/media-proxy/Dockerfile
+++ b/media-proxy/Dockerfile
@@ -189,7 +189,7 @@ FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME} as mesh-agent
 LABEL maintainer="milosz.linkiewicz@intel.com"
 LABEL org.opencontainers.image.title="Media Communications Mesh Mesh Agent"
 LABEL org.opencontainers.image.description="Media Communications Mesh Mesh Agent application. Ubuntu 22.04 Docker Container Release Image"
-LABEL org.opencontainers.image.version="1.0.0"
+LABEL org.opencontainers.image.version="25.03"
 LABEL org.opencontainers.image.vendor="Intel Corporation"
 LABEL org.opencontainers.image.licenses="BSD 3-Clause License"
 
--- a/media-proxy/Dockerfile
+++ b/media-proxy/Dockerfile
@@ -135,7 +135,7 @@ RUN ./build.sh "${PREFIX_DIR}" && \
 
 ARG IMAGE_CACHE_REGISTRY
 ARG IMAGE_NAME
-FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME}
+FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME} as media-proxy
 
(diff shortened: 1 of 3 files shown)


ko80:
@Magda-Pytel Can you provide a good description?

media-proxy/Dockerfile
```diff
@@ -182,3 +182,52 @@ SHELL ["/bin/bash", "-c"]
...
 
 HEALTHCHECK --interval=30s --timeout=5s CMD ps aux | grep "media_proxy" || exit 1
+
+
+FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME} as mesh-agent
+
+LABEL maintainer="milosz.linkiewicz@intel.com"
+LABEL org.opencontainers.image.title="Media Communications Mesh Media Proxy Agent"
+LABEL org.opencontainers.image.description="Media Communications Mesh Media Proxy Agent application. Ubuntu 22.04 Docker Container Release Image"
+LABEL org.opencontainers.image.version="1.0.0"
```
MateuszGrabuszynski, line 192: I know we use `1.0.0` as a value everywhere in the repository, but this may be a good starting point for a discussion, whether we should use the versioning matching the repository one.
https://github.com/opencontainers/image-spec/blob/main/annotations.md#:~:text=The%20version%20MAY%20match%20a%20label%20or%20tag%20in%20the%20source%20code%20repository

media-proxy/Dockerfile
```diff
@@ -182,3 +182,52 @@ SHELL ["/bin/bash", "-c"]
...
+USER mcm
+WORKDIR /opt/mcm/
+
+EXPOSE 8100/tcp 50051/tcp
+
+CMD ["-h"]
+SHELL ["/bin/bash", "-c"]
+ENTRYPOINT ["/usr/local/bin/mesh-agent"]
+
+HEALTHCHECK --interval=30s --timeout=5s CMD ps aux | grep "mesh-agent" || exit 1
```
MateuszGrabuszynski, line 233: I am worried about this health check not returning `1` ever. Would it really work? `xyz` is just a random string, not an existing/running program, and the exit code is still `0`. It is because `grep` is shown on the list as a matching program.
image

media-proxy/Dockerfile
```diff
@@ -182,3 +182,52 @@ SHELL ["/bin/bash", "-c"]
 ENTRYPOINT ["/usr/local/bin/media_proxy"]
 
 HEALTHCHECK --interval=30s --timeout=5s CMD ps aux | grep "media_proxy" || exit 1
+
+
+FROM ${IMAGE_CACHE_REGISTRY}/${IMAGE_NAME} as mesh-agent
+
+LABEL maintainer="milosz.linkiewicz@intel.com"
+LABEL org.opencontainers.image.title="Media Communications Mesh Media Proxy Agent"
+LABEL org.opencontainers.image.description="Media Communications Mesh Media Proxy Agent application. Ubuntu 22.04 Docker Container Release Image"
```
MateuszGrabuszynski, line 191: It sounds very bad with two meshes next to each other, perhaps we can come up with something even better?
```suggestion
LABEL org.opencontainers.image.title="Media Communications Mesh Mesh Agent"
LABEL org.opencontainers.image.description="Media Communications Mesh Mesh Agent application. Ubuntu 22.04 Docker Container Release Image"
```

MateuszGrabuszynski (changes requested):
A few small remarks added. Most important: Please ensure the health check is running fine.

media-proxy/Dockerfile
ko80: @Magda-Pytel Consider the following names
* Agent -> Media Communications Mesh Agent
* Proxy -> Media Communications Mesh Media Proxy

There is no such SW component like `Media Proxy Agent` in MCM.

media-proxy/Dockerfile
MateuszGrabuszynski: Just came up to my mind
```suggestion
HEALTHCHECK --interval=30s --timeout=5s CMD if(( $(ps aux | grep "mesh-agent" | wc -l) != 2 )); then exit 1; fi
```

media-proxy/Dockerfile
Magda-Pytel: So do you suggest to change to 1.1.0? :)

media-proxy/Dockerfile
MateuszGrabuszynski: I think Konstantin's version is much clearer. Mine came from confusion.

media-proxy/Dockerfile
MateuszGrabuszynski: I am not sure if we should not use `v25.03` (or `25.03`) as the version? Like it was versioned in the same manner as the repository. Sorry to put you on spot @Mionsz but you seem to be the most familiarized with that topic, could you provide any guidance?

MateuszGrabuszynski: See:
https://github.com/search?q=repo%3AOpenVisualCloud%2FIntel-Tiber-Broadcast-Suite%20org.opencontainers.image.version%3D&type=code
VS
https://github.com/search?q=repo%3AOpenVisualCloud%2FMedia-Communications-Mesh%20org.opencontainers.image.version%3D&type=code

MateuszGrabuszynski (approved):
lgtm 🚀

Mionsz (approved):

PR #343, opened 2025-02-28
Add building artifacts for Ubuntu build

zLukas:
ADD: 
* build artifacts for ubuntu build


KarolinaPomian (dismissed):
lgtm

ko80 (approved):

.github/workflows/ubuntu-build.yml
```diff
@@ -94,6 +94,11 @@ jobs:
         ./configure-ffmpeg.sh "7.0" --disable-doc --disable-debug && \
         ./build-ffmpeg.sh "7.0"
 
+    - name: 'upload media-proxy and mcm binaries'
+      uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
+      with:
+        name: mcm-build
+        path: '${{ env.BUILD_DIR }}/*'
```
Mionsz, line 101: This path includes literally every project that Media-Communications-Mesh is dependent on. Including full repositories clones, build artifacts etc. Please send only the `mcm` part and only the binary files.

Mionsz (changes requested):
@zLukas if you can please include the retention tam in the process so that it is fixed to specific number - like probably 30 or 45 days of retention.

zLukas:
> @zLukas if you can please include the retention tam in the process so that it is fixed to specific number - like probably 30 or 45 days of retention.

by default Github set 90 day retention time, so nothing to change to meet this requirement:
https://docs.github.com/en/enterprise-server@3.10/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization

Mionsz (approved):

ko80 (approved):

PR #349, opened 2025-03-12
Add: Improvements to TestApp (validation)

MateuszGrabuszynski:
- [ ] Ctrl+C graceful shutdown - inform media_proxy instance about deactivating the tx/rx app (SIGINT handler)
- [ ] Pacing is hardcoded at 25 fps at the moment https://github.com/OpenVisualCloud/Media-Communications-Mesh/blob/9552a01919dd9a9928a1fc737a427273d20cc991/tests/tools/TestApp/src/mcm.c#L52 it would be great if we improve it by making it editable by the user (parametrized)
- [ ] Looping - allow the user to read the file as many times as needed; either make it loopable endlessly, or allow to put the exact number of loops

Connected to: SDBQ-2080

### DIFF
--- a/tests/tools/TestApp/src/mcm.c
+++ b/tests/tools/TestApp/src/mcm.c
@@ -26,6 +26,8 @@ int mcm_send_video_frames(MeshConnection *connection, const char *filename) {
 
     unsigned int frame_num = 0;
     size_t read_size = 1;
+    float one_sec = 1000000; /* microseconds (usec) */
+    float requested_fps = 25.0; /* FIXME: Read requested fps value */
     while (1) {
 
         /* Ask the mesh to allocate a shared memory buffer for user data */
@@ -47,9 +49,7 @@ int mcm_send_video_frames(MeshConnection *connection, const char *filename) {
             goto close_file;
         }
 
-        /* Temporary implementation for pacing */
-        /* TODO: Implement pacing calculation */
-        usleep(40000);
+        usleep(one_sec/requested_fps);
     }
     LOG("[TX] data sent successfully");
 close_file:


MateuszGrabuszynski:
Checking the changes, will tick the list above when checked.

MateuszGrabuszynski (commented):
I cannot approve myself, as the last commiter, but it lgtm ;)

tests/tools/TestApp/rx_app.c
```diff
@@ -49,18 +54,33 @@ int main(int argc, char *argv[]) {
...
     free(conn_cfg);
-    return err;
+    exit(sig);
+}
+
+void setup_signal_handler() {
+    sa.sa_handler = handle_sigint;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_flags = 0;
+    sigaction(SIGINT, &sa, NULL);
```
ko80, line 85: Should handle both SIGINT and SIGTERM.

ko80: Test that by sending the following commands:
```sh
kill -SIGINT <PID>
kill -SIGTERM <PID>
```

tests/tools/TestApp/tx_app.c
```diff
@@ -51,20 +57,36 @@ int main(int argc, char **argv) {
...
+    }
+    if (client) {
+        LOG("[TX] Shuting down client");
+        mesh_delete_client(&client);
+    }
     free(client_cfg);
     free(conn_cfg);
-    return err;
+    exit(sig);
 }
```
ko80, line 84: This could crash the app if `mesh_delete_connection()` is called while `mesh_get_buf()` or `mesh_put_buf()` are in progress in the main thread.

Rework this all to raise an atomic flag named `shutdown_requested` and check it periodically in the main loop of sending or receiving data. The graceful shutdown concept means that your app exits correctly from the `main()` function, not from the sigint handler.

tests/tools/TestApp/rx_app.c
zLukas: > Should handle both SIGINT and SIGTERM.

done

tests/tools/TestApp/tx_app.c
```diff
@@ -51,20 +65,34 @@ int main(int argc, char **argv) {
...
         goto safe_exit;
     }
 
-    /* Open file and send its contents */
-
-    err = mcm_send_video_frames(connection, video_file);
-    LOG("[TX] Shuting down connection and client");
+    /* Open file and send its contents in loop*/
+    while(1){
+        err = mcm_send_video_frames(connection, video_file);
```
ko80, line 73: Can you confirm `mcm_send_video_frames()` sending a long 2 minutes file exits when the termination signal occurs?

MateuszGrabuszynski: It does work fine, even after a few minutes.

tests/tools/TestApp/src/mcm.c
```diff
@@ -46,10 +53,7 @@ int mcm_send_video_frames(MeshConnection *connection, const char *filename) {
             LOG("[TX] Failed to put buffer: %s (%d)", mesh_err2str(err), err);
             goto close_file;
         }
-
-        /* Temporary implementation for pacing */
-        /* TODO: Implement pacing calculation */
-        usleep(40000);
+        usleep(sleep_us);
```
tszumski, line 56: You need to measure how much time has passed between line "mesh_get_buffer()" and "mesh_put_buffer()"
Then you can execute usleep with adjusted time::
usleep(sleep_us - elapsed_time);

ko80: Yeah, the pacing has to be linear.

MateuszGrabuszynski: As we discussed on the call - this is not a reference implementation, thus it may not be precise. Tried adding the precision from the sample app breaks other things. We do not have much more time to work on making it any better at the moment.

ko80 (dismissed):

tszumski (approved):

tests/tools/TestApp/src/mcm.c
MateuszGrabuszynski: After a few other fixes, the pacing (copied from sample app) works as well. ;)

tszumski (approved):

MateuszGrabuszynski (commented):
Works properly on my end :)

tests/tools/TestApp/src/mcm.c
tszumski: Thanks, looks good!

tests/tools/TestApp/src/mcm.c
```diff
@@ -91,6 +112,10 @@ void read_data_in_loop(MeshConnection *connection, const char *filename) {
             break;
         }
         LOG("[RX] Frame: %d", ++frame);
+        if (graceful_shutdown && graceful_shutdown() != 0 ) {
```
ko80, line 115: Why don't you simply make `shutdown` visible in all source files? The fact that you define a function to return the flag and then pass the pointer to this function to other functions makes the flow unnecessary overcomplicated.

I suggest to rework it and just check the `shutdown` flag here.

ko80 (approved):

PR #351, opened 2025-03-13
Update integrity lib with method using bigger output files

andremiszcz:
Update integrity lib with function for checking bigger files than source one.

### DIFF
--- a/tests/validation/Engine/integrity.py
+++ b/tests/validation/Engine/integrity.py
@@ -60,6 +60,36 @@ def check_chunk_integrity(src_chunk_sums, out_chunk_sums, expected_frame_percent
     # only return true if no more frames left in out frames list, else fail
     return out_index == len(out_chunk_sums)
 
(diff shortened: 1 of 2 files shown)


tests/validation/Engine/integrity.py
```diff
@@ -22,30 +22,43 @@ def calculate_chunk_hashes(file_url: str, chunk_size: int) -> list:
...
+            first_shift = False
+            continue
+        # if first time shift, shift source and continue
+        if first_shift:
+            src_index += 1
+            continue
+        # not a first shift, corrupted in the middle
+        return False
+    # only return true if no more frames left in out frames list, else fail
+    return out_index == len(out_chunk_sums)
```
zLukas, line 61: Return false in line 59 will always end the function is no 'if' statement was matched, resulting, line 61 never executing.

MateuszGrabuszynski: Only if the while loop entry condition is not true, then line 61 is executed.

zLukas: ok, I got it now

zLukas (approved):

PR #353, opened 2025-03-18
Fix issues appeared in version 25.03-rc4

ko80:
This pull request includes the following commits
* Add version info to Agent, Media Proxy, and SDK
* Increase string storage for JSON config in FFmpeg MCM plugin
* Improve shutdown flow in Media Proxy
* Add workaround to avoid first frames losses
* Add support for conn creation delay in FFmpeg MCM plugin
* Fix last Tx packet len in FFmpeg MCM plugin audio device
* Add support for ST2110 payload type (96..127)
* Fix build system to support version info in separate SW components


moleksy (approved):
Looks good to me.

PR #357, opened 2025-03-20
Val: Prepare a tool to check delays between frames based on log output

MateuszGrabuszynski:
## Warnings
- At the moment this code can produce misleading results with wrong data assigned to wrong csv header (header to data shifts). That is why there is always a `filename-` before the frame number.
- When more than a single result from a single timestamp, there is no `CSV_DELIMITER` printed for the latter results.

## To-dos
- [ ] Use pandas' DataFrame instead of current approach with static printing
- [ ] Deal with a possibility to have more than a single result from a single timestamp ("zero" delay)
- [ ] Allow changing keyword with parameters
- [ ] Draw a final graph with delays (properly shifted based on timestamps)

### DIFF
--- a/tests/validation/Engine/log-timing-checker.py
+++ b/tests/validation/Engine/log-timing-checker.py
@@ -1,0 +1,74 @@
+import argparse
+from datetime import datetime, timezone
+
+EXECUTION_TIME = datetime.now(timezone.utc) # used for default year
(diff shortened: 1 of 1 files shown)


MateuszGrabuszynski:
Low priority, but after changes, may be used in the near future (a few months).

PR #359, opened 2025-03-23
Fix connection shutdown handling flow

ko80:
This pull request includes the following commits:
* Implement lock-free hotpath in conn and multipoint group.
* Fix FFmpeg MCM plugin Rx connection shutdown related issues.

### DIFF
--- a/sdk/src/mesh_conn.cc
+++ b/sdk/src/mesh_conn.cc
@@ -970,10 +970,11 @@ int ConnectionContext::shutdown()
                         }
 
                         err = buf_ctx->enqueue(500);
-                        if (err)
-                            break;
 
                         delete buf_ctx;
+
+                        if (err)
+                            break;
                     }
                 });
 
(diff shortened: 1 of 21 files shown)


sdk/src/mesh_conn.cc
```diff
@@ -935,23 +937,57 @@ int ConnectionContext::shutdown()
...
+                        if (!buf_ctx)
+                            break;
+                    
+                        int err = buf_ctx->dequeue(500);
+                        if (err) {
+                            delete buf_ctx;
+                            break;
+                        }
+
+                        err = buf_ctx->enqueue(500);
```
moleksy, line 972: Shouldn't we delete buf_ctx here too?

ko80: Wow, perfect catch, thank you! Will fix it shortly.

ko80: Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/359/commits/72189c36a74c58f0767b8408d52b1fd3b3cfe91d

tszumski (approved):

moleksy (approved):
Good job!

PR #364, opened 2025-03-26
FIX: More dashboard and trivy relatred fixes

Mionsz:
FIX: More dashboard and trivy relatred fixes.
Add: Missing pip version pinnings
Add: Missing hash version hooking in github actions.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,11 @@
+# Usability Notice
+
+The data contained within the `tests` folder MUST NOT be used as a reference or guide for any external development.
+
+Please note that this folder may contain code or instructions that:
+- Are not up to date with the main branch or any other branch (even when present on that branch),
+- Use undocumented or deprecated features,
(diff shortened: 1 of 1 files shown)


scripts/setup_build_env.sh
```diff
@@ -143,7 +143,7 @@ function install_yum_package_dependencies()
         texinfo \
         wget \
         zlib-devel && \
-    python3 -m pip install meson ninja && \
+    python3 -m pip install meson~=0.61.2 ninja~=1.11.1 && \
```
MateuszGrabuszynski, line 146: The newest meson version that complies with that condition is from 2022. Is it a strict requirement, that we use pre 1.0 meson?

MateuszGrabuszynski (commented):
It is a good practice to add those package versions to ensure stable execution conditions. However, we should keep in mind, they all shall be less than 1 year old, unless strictly necessary.

MateuszGrabuszynski (dismissed):
LGTM

PR #367, opened 2025-03-27
Fix data plane atomic pointer implementation

ko80:
* Rework the data plane atomic pointer class to avoid bypassing the lock after timeout.
* Use atomic dual purpose command/state storage to pass new pointer value.
* Implement lock/unlock functionality to instantiate a critical section.
* Add protection in the Local Rx class to avoid passing a null buffer or an empty buffer.

### DIFF
--- a/media-proxy/src/mesh/sync.cc
+++ b/media-proxy/src/mesh/sync.cc
@@ -8,39 +8,56 @@
 
 namespace mesh::sync {
 
-uint64_t DataplaneAtomicUint64::load()
+static const uint64_t STATE_UNLOCKED = 0;
+static const uint64_t STATE_LOCKED   = 2;
+static const uint64_t FLAG_NEW_VALUE = 1;
+
+void * DataplaneAtomicPtr::load()
 {
-    return current.load(std::memory_order_acquire);
+    return (void *)current.load(std::memory_order_acquire);
 }
 
-void DataplaneAtomicUint64::store_wait(uint64_t new_value,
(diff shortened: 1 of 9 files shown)


tszumski (approved):

media-proxy/src/mesh/sync.cc
```diff
@@ -8,39 +8,56 @@
...
 
-void DataplaneAtomicUint64::store_wait(uint64_t new_value,
-                                       std::chrono::milliseconds timeout_ms)
+void DataplaneAtomicPtr::store_wait(void *new_ptr)
 {
     std::lock_guard<std::mutex> lk(mx);
 
-    next.store(new_value, std::memory_order_release);
-
-    auto ctx = context::WithTimeout(context::Background(),
```
moleksy, line 23: Since we don't have timeout anymore, couldn't there be the case that if the hotpath never unlocks, DataplaneAtomicPtr::store_wait will block indefinitely?

moleksy: Resolved offline.

moleksy (approved):
Looks good, nice solution!

media-proxy/src/mesh/sync.cc
ko80: Yes, it will block infinitely the same way as the regular std::mutex.
The hotpath should never block forever. The processing flow should be designed and implemented the way to avoid it.

PR #388, opened 2025-04-17
Add timeout handling in RDMA buffer sending

ko80:
The ep_send_buf() function keeps running a loop forever. Avoid this by checking the timeout.


tszumski (dismissed):

media-proxy/src/libfabric_ep.c
```diff
@@ -210,6 +215,16 @@ int ep_send_buf(ep_ctx_t* ep_ctx, void* buf, size_t buf_size) {
...
             (void)fi_cq_read(ep_ctx->cq_ctx.cq, &cq_entry, 1); // Drain CQ
         }
+
+        // Get the current time and calculate the elapsed time
+        clock_gettime(CLOCK_MONOTONIC, &current_time);
+        elapsed_time_ms = (current_time.tv_sec - start_time.tv_sec) * 1000 +
+                        (current_time.tv_nsec - start_time.tv_nsec) / 1000000;
+
+        // Check if the elapsed time exceeds the timeout interval
+        if (elapsed_time_ms > 100)
```
moleksy, line 225: If we have succesfull send and timeout expires we still get error -ETIMEDOUT, perhaps we should move it inside the if (ret == -EAGAIN) branch?

ko80: Oh, good catch. Fixed in https://github.com/OpenVisualCloud/Media-Communications-Mesh/pull/388/commits/f80b82b6cf1a258d7f1dc294cb36092c37534717

moleksy (approved):

Sakoram (approved):
There are a few different ways to approach this issue.
1. we could use fi_cq_sread to block while waiting for completion event, return if timeout was reached or try to send when event is received, with assumption that after that the fi_send will be successful
2. similarly we could use fi_cntr_wait to wait till tx counter is increased, and after that try to send again.

but I think this solution is alright

